Issue #52, opened 2021-07-15
Amcrest AD410 is supported


andrew-d:
## Version of the custom_component
0.9.2

## Configuration
n/a

## Describe the bug
I wanted to report that the [Amcrest AD410](https://amcrest.com/smarthome-4-megapixel-wireless-doorbell-security-camera-2560-x-1920-wifi-doorbell-camera-ip65-weatherproof-two-way-audio-ad410.html) video doorbell appears to be supported with this integration. Some brief notes in case anyone else has this:

* Pressing the doorbell appears to fire the `alarm_local` binary sensor - e.g. `binary_sensor.front_doorbell_alarm_local` here.
* The camera shows up in Home Assistant and works fine, including both SD/HD streams and sound.
* I haven't figured out how to get the "human detection" working yet; `binary_sensor.front_doorbell_smart_motion_human` doesn't appear to fire, but I'll play with it a bit more.

I'm happy to do any debugging/etc. if any of the maintainers want - but I figured I'd start with the fact that it works 🎉  Thanks to everyone that's contributed!

## Debug log
n/a

rroller:
I just released 0.9.3 where I treat the AD410 as a doorbell instead of just a camera. Can you try it out to make sure things works as expected? If not I'll need to revert the change.

andrew-d:
I updated, then removed and re-added the integration; here's what I tested:

* `binary_sensor.front_doorbell_alarm_local` goes unsafe/safe when I ring the doorbell
* The SD/HD video streams work, including sound
* `binary_sensor.front_doorbell_call_no_answered` goes from "motion detected" to "cleared" when I press the doorbell

I do see this message in the logs that I don't remember seeing before, but unfortunately I can't confirm/deny that since I don't think I have older logs:

```
2021-07-17 15:28:39 ERROR (Thread-4) [custom_components.dahua] Failed to handle message, error: 'NoneType' object is not iterable, Line: 100
```

Also, with the debug logs on, here's what I think happens when a human is detected by the doorbell (pretty-printed to be easier to read, some numbers redacted with `1` because I'm unsure if they're sensitive):

<details><summary>Click to expand</summary>

```
2021-07-17 15:37:16 DEBUG (Thread-4) [custom_components.dahua] VTO Data received: {
 'Action': 'Start',
 'Code': 'CrossRegionDetection',
 'Data': {'Action': 'Appear',
          'CfgRuleId': 1,
          'Class': 'Normal',
          'CountInGroup': 0,
[...]

PR #66, opened 2021-08-03
vto: don't attempt to iterate over an empty table

andrew-d:
This fixes the bug described in #52 - the `table` variable is None and thus the for loop throws an exception. See here: https://github.com/rroller/dahua/blob/9910d0d/custom_components/dahua/vto.py#L256


rroller (approved):

rroller:
Thank you!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.13

PR #79, opened 2021-08-31
Support for Lorex E891AB NVR

brianegge:
I have an eight channel Lorex NVR (along with various Amcrest/Loryta) cams. Unfortunately, four of the lorex cams are still using the PoE on the NVR. 

Lorex is returning "31" for the model, but has the model in the unqueSerial field. Also, the day/night mode access through the NVR uses a different method. The Dahua method works on the direct connected NVRs, but not through the Lorex firmware.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -184,7 +184,7 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                     data.update(response)
 
                 device_type = data.get("deviceType", None)
-                if device_type == "IP Camera" or device_type is None:
+                if device_type in ["IP Camera", "31"] or device_type is None:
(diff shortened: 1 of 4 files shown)


custom_components/dahua/__init__.py
```diff
@@ -184,7 +184,7 @@ async def _async_update_data(self):
                     data.update(response)
 
                 device_type = data.get("deviceType", None)
-                if device_type == "IP Camera" or device_type is None:
+                if device_type in ["IP Camera", "31"] or device_type is None:
```
rroller, line 187: Do you mind leaving a comment about "31". That will help for future maintenance.

rroller (approved):
Wow! This is awesome, thanks for this!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.922

PR #180, opened 2022-03-25
Generic RTSP Camera support update client.py

dmr1987:
Generic RTSP Camera support

1. Added exception with default values.
2. Return value for _get_max_extra_streams()_ changed from **2** to **3**
3. Return value for _get_rtsp_stream_url()_ changed for **subtype==3**
4. For clean RTSP during setup camera value of Address must provide like _192.168.17.12/stream1_
5. Entity for clean RTSP-stream is **sub_3**, for example _rtsp_camera_3_sub_3_, _outdoor_camera_sub_3_

### DIFF
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -6,6 +6,7 @@ import aiohttp
 import async_timeout
 
 from .digest import DigestAuth
+from hashlib import md5
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
@@ -53,6 +54,13 @@ class DahuaClient:
             channel,
             subtype,
         )
+        if subtype==3 :
+            url = "rtsp://{0}:{1}@{2}".format(
+                self._username,
+                self._password,
+                self._address,
+            )
+
         return url
 
     async def async_get_snapshot(self, channel_number: int) -> bytes:
@@ -77,7 +85,13 @@ class DahuaClient:
         updateSerial=IPC-HDW5830R-Z
         updateSerialCloudUpgrade=IPC-HDW5830R-Z:07:01:08:70:52:00:09:0E:03:00:04:8F0:00:00:00:00:00:02:00:00:600
         """
-        return await self.get("/cgi-bin/magicBox.cgi?action=getSystemInfo")
+        not_hashed_id = "{0}_{1}_{2}_{3}".format(self._address, self._rtsp_port, self._username, self._password)
+        unique_cam_id = md5((not_hashed_id).encode('UTF-8')).hexdigest()
+        try:
+            return await self.get("/cgi-bin/magicBox.cgi?action=getSystemInfo")
+        except aiohttp.ClientResponseError as e:
+            return {"serialNumber":unique_cam_id}
+
 
     async def get_device_type(self) -> dict:
         """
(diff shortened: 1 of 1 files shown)


rroller:
Thanks! I'll check this out soon. Been busy and am behind on issues and prs. Will try for tomorrow

custom_components/dahua/client.py
```diff
@@ -77,7 +85,13 @@ async def async_get_system_info(self) -> dict:
         updateSerial=IPC-HDW5830R-Z
         updateSerialCloudUpgrade=IPC-HDW5830R-Z:07:01:08:70:52:00:09:0E:03:00:04:8F0:00:00:00:00:00:02:00:00:600
         """
-        return await self.get("/cgi-bin/magicBox.cgi?action=getSystemInfo")
+        not_hashed_id = "{0}_{1}_{2}_{3}".format(self._address, self._rtsp_port, self._username, self._password)
+        unique_cam_id = md5((not_hashed_id).encode('UTF-8')).hexdigest()
```
rroller, line 89: Seems this work should only be done in the "except" case. Otherwise it's done and thrown away for nothing.

custom_components/dahua/client.py
```diff
@@ -87,22 +101,36 @@ async def get_device_type(self) -> dict:
...
 
     async def get_machine_name(self) -> dict:
         """ get_machine_name returns the device name. Example response: name=FrontDoorCam """
-        return await self.get("/cgi-bin/magicBox.cgi?action=getMachineName")
+        not_hashed_id = "{0}_{1}_{2}_{3}".format(self._address, self._rtsp_port, self._username, self._password)
+        unique_cam_id = md5((not_hashed_id).encode('UTF-8')).hexdigest()
+        try:
+            return await self.get("/cgi-bin/magicBox.cgi?action=getMachineName")
+        except aiohttp.ClientResponseError as e:
+            return {"name":unique_cam_id}
```
rroller, line 126: Can you leave a comment here what this is. Any context would help

rroller:
released in https://github.com/rroller/dahua/releases/tag/0.9.40

custom_components/dahua/client.py
dmr1987: Dahua camera return unique identifier by "/cgi-bin/magicBox.cgi?action=getMachineName". For RTSP camera need to generate unique id ad use this id everywhere for identify objects. File core.device_registry provide json information like this

 {
                "config_entries": [
                    "374e8200431dc735791ae0f63556d781",
                    "100389553d3d6524d8c73dde2e474d00"
                ],
                "connections": [
                    [
                        "mac",
                        "00:30:6f:83:d1:21"
                    ]
                ],
                "identifiers": [
                    [
                        "keenetic_ndms2",
                        "00:30:6f:83:d1:21"
                    ],
                    [
                        "dahua",
                        "8026a73660b0371a68cfe9169ca6830d"
                    ]
                ],
                "manufacturer": "Dahua",
                "model": "Generic RTSP",
                "name": "RTSP_Camera_1",
                "sw_version": "1.0",
                "hw_version": null,
                "entry_type": null,
                "id": "32777735d851d72aa73a8c6e65b33597",
                "via_device_id": null,
                "area_id": "ulitsa",
                "name_by_user": null,
                "disabled_by": null,
                "configuration_url": "http://192.168.17.11:554/cam0_0"
            },

custom_components/dahua/client.py
dmr1987: Yes it is. For RTSP need to make json structure that can be used by "class DahuaDataUpdateCoordinator" in file "__init__.py"

Issue #61, opened 2021-07-24
Missing binary sensor - button pressed


djxela:
I have VTO 2211G-P everything works good but I dont see binary sensor which reports when butt is pressed.  In configuration I have checked  option "binary sensor". Do you know what is the problem?

rroller:
It should look like this...

![image](https://user-images.githubusercontent.com/445655/126875781-2df7a43e-64e7-40a5-8ca5-f807d3055ca7.png)

djxela:
I have only these:
![46F9075A-E8B9-4762-ACD5-8E16E9A4E9B1](https://user-images.githubusercontent.com/36244426/126877819-e7f83f92-120e-415e-bb6b-4db8fa4e8ea3.jpeg)

djxela:
Also after upgrade I dont have new entity: switch to  open the door

djxela:
Should it appear automatically after configuration finisz or I should configure it manually?

djxela:
This from dahuavto2mqtt works:
![909504FC-C92A-4BD6-A2AC-31292E01ECB6](https://user-images.githubusercontent.com/36244426/126879071-90ca4ee2-18c2-4d92-8746-20ee98c4137c.jpeg)

rroller:
It should just show up

djxela:
So what could be the problem

papinist:
I have the same problem - missing binary sensor about button press. I have a VTO3211D-P4-S2.
Btw your integration works perfect with this VTO :)

nibblerrick:
I have a VTO3211D, too, but the P2. The binary sensor is there, so that's not a problem. But is has two buttons, so I agree, it would be nice to have two sensors, one for each button. Otherwise the integrations works fine with the VTO3211D-P2 and a VTO2202F!

papinist:
Ok I found the binary sensor - I was confused that it does not appear in the device's entity list, but when creating an automation, it is present.
However it would be useful to have distinct sensors for each button.
Thanks!

rroller:
@papinist @nibblerrick  you can always use the event bus instead of the binary sensors. See https://github.com/rroller/dahua/blob/main/README.md#events

All button presses, etc are fired on the HA event bus and you can create automations from that. The binary sensors are just kind of nice-to-have helpers for simpler cases.

For adding the other buttons... can enable debug logging, press the other button, and paste the event data here? I don't have a unit with two or more buttons so I'll need the event data to try and add it.

PR #198, opened 2022-06-26
Detect VTO-2211 (newer firmware?) as doorbell

pedro042:
Hi, my VTO-2211-P was not detected as doorbell. Device name is DH-VTO2211G-P (it doesn't match VTO-* DHI-* etc.)
Firmware version is 4.511.0000000.0.R,build:2021-11-16.
This is probably related to https://github.com/rroller/dahua/issues/61
Now it's detected as doorbell and the binary sensor works.
Thank you very much for this integration


rroller:
Thank you!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.41

PR #212, opened 2022-08-14
Add `ObjectRemovalDetection` and `ObjectPlacementDetection`

mdegat01:
I have an IPC-T5442T-ZE running on firmware V2.840.15OG008.0.R (Build Date: 2022-02-18). I wanted to use the "Smart Object Detection" feature via this integration to detect when packages were left and removed. Initially I tried to do this by selecting the `TakenAwayDetection` and `LeftDetection` events but that didn't do anything. After some digging I realized its because my camera fires events with codes `ObjectRemovalDetection` and `ObjectPlacementDetection` here.

I'm not sure if this changed in newer firmware or different models do different things so I didn't modify the existing events. But it would be good if these two could get added to the list so folks like myself could use them.

### DIFF
--- a/custom_components/dahua/config_flow.py
+++ b/custom_components/dahua/config_flow.py
@@ -66,6 +66,8 @@ ALL_EVENTS = ["VideoMotion",
               "CrowdDetection",
               "FireWarning",
               "FireWarningInfo",
+              "ObjectPlacementDetection",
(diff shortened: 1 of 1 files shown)


rroller:
Thanks!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.41

PR #260, opened 2023-02-18
Reduce error message verbosity for simple connection errors

redstone99:
A bit of cleanup around log messages and error handling when connection to a camera fails (e.g., the camera isn't plugged in). The component used to add a lot of long stack trace and error messages if a connection error happened. Now it's more like a single warning line saying the connection failed (with host & port).


rroller:
Thanks

rroller:
Released in 0.9.45 - https://github.com/rroller/dahua/releases/tag/0.9.45

PR #286, opened 2023-06-21
fix MaxExtraStream spelling

si458:
the checker is looking for the wrong spelling
checked my XVR (BNC) and its showing 1 extra stream `table.MaxExtraStream=1`
checked my NVR (IP) which returns 2 extra stream `table.MaxExtraStream=2`

also fixes my other issue where the XVR keeps returning 3 streams
because it cant find the value so just assumes its 2 extra streams when really it should be 1


rroller (approved):

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.51

PR #301, opened 2023-09-09
VTO buffer and keepalive fix

urbim:
I had some issues running this integration with DHI-VTO3311Q-WP. After some time, the integration stopped working and I had to restart homeassistant to fix it.

After some debugging, it turns out that asyncio does not forward the entire message to the data_received function if the message is too large. In turn this sometimes messes up the future messages and the integration stops working. To remedy that, I added a buffer to the data_received function.

During debugging I also noticed that data_handlers is growing due to keep_alive handlers being added constantly. To fix that I added code to remove the handler after response from keep-alive has been received. I don't think this has anything to do with the above issue, however I still think it should be fixed.

And for the last change - I added DHI-VTO3311Q-WP to the list of supported devices, since I had no further issues running the integration with the above fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Brand | 2 Megapixels | 4 Megapixels | 5 Megapixels | 8 Megapixels
 | *Dahua* |
 | | DHI-VTO2202F-P |
 | | DHI-VTO2211G-P |
+| | DHI-VTO3311Q-WP |
 | *IMOU* |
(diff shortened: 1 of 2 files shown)


rroller:
Thank you! Merging, releasing, and testing it. Yes I should test then release, but I don't have my local HA setup currently :-)

rroller:
https://github.com/rroller/dahua/releases/tag/0.9.48

Issue #263, opened 2023-03-10
Lorex Floodlight support requested


josetann:
It's another rebadged Dahua. To test, I replaced

       return self.model.upper().startswith("ASH26")

with

       return self.model.upper().startswith("V261LC")

and it works fine. Well, I can't change the brightness, but I can turn the floodlight on and off as expected.

secabeen:
This looks great, thank you!  I think a general pull request to add all of the Lorex Model numbers in the same way as Amcrest devices would be a good idea.  I can help if we can put together the table of devices needed.

zios0007:
This is in, and it is your patch.

`custom_components/dahua/__init__.py`:

```python
return m.startswith("ASH26") or "L26N" in m or "L46N" in m \
    or m.startswith("V261LC") or m.startswith("W452ASD")
```

`V261LC` is there, so the Lorex floodlight is recognised without anyone having to edit the file.

@josetann thank you for writing out the exact steps rather than just reporting it. That is why it was straightforward to land.

@secabeen on the port: the integration asks for the HTTP port when you add the device, so if yours does not listen on 80, put the port it does use in that field. If it serves nothing over HTTP at all, that is a different problem and worth its own issue.

Worth noting the wider direction this was part of: deciding capabilities by matching model strings is being unwound in favour of asking the device, because a list of names can never cover the OEM rebrands (#676).

@rroller closeable.

PR #313, opened 2023-11-07
Add support for L46N floodlights and possibly the Lorex equivalent

cchamilt:
Hi, Lightingv2 doesn't trigger the floodlights in the later Dahua model L46N.  After some exploration and comparison to DMSS call status, it appears that the manual floodlight trigger is only available through the coaxial API.  This also requires the table.FloodlightMode.mode to be set to manual or at least not motion detection to stay on.  I tried to add it along side the Amcrest work assuming it still works for the older model.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -253,8 +253,8 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                 is_doorbell = self.is_doorbell()
                 _LOGGER.info("Device is a doorbell=%s", is_doorbell)
 
-                is_amcrest_flood_light = self.is_amcrest_flood_light()
-                _LOGGER.info("Device is an Amcrest floodlight=%s", is_amcrest_flood_light)
+                is_flood_light = self.is_flood_light()
+                _LOGGER.info("Device is a floodlight=%s", is_flood_light)
 
                 try:
                     await self.client.async_get_config_lighting(self._channel, self._profile_mode)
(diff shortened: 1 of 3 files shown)


rroller (approved):

rroller:
I just looked at your profile, we are in the same city :D

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.50

josetann:
Any possibility to add in support for "V261LC" ? I reported earlier how I got it to work, looks like you went the long (correct) way around, but maybe just adding V261LC would work with your fix?

My prior post: https://github.com/rroller/dahua/issues/263

cchamilt:
Hi, I added in detection with m.startswith("V261LC"), but I didn't assign it the Coaxial API and kept the original LightingV2 calls.  It may need the Coaxial API as well.  I guess it might be a firmware version specific change.

josetann:
> Hi, I added in detection with m.startswith("V261LC"), but I didn't assign it the Coaxial API and kept the original LightingV2 calls. It may need the Coaxial API as well. I guess it might be a firmware version specific change.

Ah, I must have skipped that particular commit. I bit the bullet and it was there as expected. Except it didn't work. When I tried to turn on or off the light, this error pops up (I had to type it out, it's possible there's a typo in there somewhere):

Failed to call service light/turn_on. 400,message='Bad Request',url=URL(http://192.168.1.78:80/cgi-bin/configManager.cgi?action=setConfig&FloodLightMode.Mode=2')

Does it look like an error in the code, or is it handled differently than the ASH26 (because in prior versions that I hacked in support, I just threw that in the same line as the ASH26 and it worked)?

cchamilt:
Hmm, the floodlight mode should be exclusive to the newer floodlights.

cchamilt:
It appears that the V261LC supports the Coaxial API but not the floodlightmode.  So they will need separated out as different floodlight types and the logic split by that instead of Coaxial API support.

cchamilt:
OK, I separated out the floodlightmode logic more clearly for the 4x floodlights.  So it shouldn't effect you now.  Please check the new PR I just made.

PR #350, opened 2024-04-27
Document support for Amcrest AD110

JackDanger:
I have an Amcrest AD110 running smoothly with this integration. The only reason this doorbell might seem to not work is because it requires a specific firmware version installed for the doorbell API to function at all (https://www.reddit.com/r/homeassistant/comments/nttciw/comment/ijat032/)

Thanks for the great work!


rroller (approved):

rroller:
Thanks!

PR #370, opened 2024-07-08
Fix day/night mode switch

hawkeye217:
I had recently updated my integrations and noticed that my day/night mode was no longer working. I traced it to a bug that was introduced in https://github.com/rroller/dahua/pull/327 - Non-empty Python strings are always truthy, so this statement always evaluates to `True`:

```
if 'NVR4108HS' or 'IPC-Color4K' in model:
```

This PR correctly searches the model string and uses an array to make it easier to add more models in the future.

### DIFF
--- a/custom_components/dahua/camera.py
+++ b/custom_components/dahua/camera.py
@@ -2,16 +2,18 @@
 from __future__ import annotations
 
 import logging
+import voluptuous as vol
 
-from homeassistant.components.camera import Camera, CameraEntityFeature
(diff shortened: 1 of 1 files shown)


rroller (approved):
Thank you! And thanks for your work on Frigate!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.59

jakeycrx:
My Dahua camera always returns a 401 with a RESTful command Is there any workaround for this?

PR #394, opened 2024-10-12
Fix hassio warnings and false True (positive) for DisableLinkage feature

jandusek4:
Fix hassio warning about blocking SSL setup `Detected blocking call to load_default_certs`  and
`setup without awaiting async_forward_entry_setup, which can cause the setup lock to be released before the setup is done`

Also fixed approach how to detect if DisableLinkage feature is present which return false True (positive) even when the feature is not present in the camera and direct cgi-bin call return 400 error.


rroller:
Thank you!

rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.61

jandusek4:
For retest, I replaced my local modified code with latest release and is is OK in logs.

Issue #398, opened 2024-10-25
Support for "Event Notifications" for Arming/Disarming


jakon89:
**Is your feature request related to a problem? Please describe.**

<img width="710" alt="Screenshot 2024-10-25 at 16 02 23" src="https://github.com/user-attachments/assets/9aa36a6e-63a4-43c5-926f-8929d27462c6">

Dahua NVRs (at least part of them) support "Event notification" boolean switch which is useful when the system is disarmed: if checked and the system is disarmed then DMSS won't get any notifications regarding any rule breakage.   

This setting is only applied when system is disarmed based on:
* Dahua camera (SD3E405DB-GNY-A-PV1) settings description:
<img width="744" alt="Screenshot 2024-10-25 at 16 27 07" src="https://github.com/user-attachments/assets/a8af2a7d-7f6a-48b2-8d10-eca3afbe6c2d">

* modal on NVR: <img width="566" alt="Screenshot 2024-10-25 at 16 06 04" src="https://github.com/user-attachments/assets/1325a28a-6aea-4d7c-91e7-5df2818048c3">

**Describe the solution you'd like**
I would like to have additional boolean switch: "Event Notifications" that controls above setting in NVR. I am not sure if this is possible but new switch should be only visible when the system is disarmed.

**Describe alternatives you've considered**
just have additional boolean switch and don't care about armed/disarmed state - the same way NVR settings work

rroller:
Thanks! I merged and released your change

PR #400, opened 2024-10-25
Add event notifications boolean switch

jakon89:
More details: https://github.com/rroller/dahua/issues/398


rroller:
Released in https://github.com/rroller/dahua/releases/tag/0.9.64

PR #405, opened 2024-10-30
Update __init__.py added support for IP8M type camera like model IP8M-2796E-AI 

alpauna:
Modified the or to add a starts with on the model looking for IP8M  in the model name.


alpauna:
Tested on my IP8M-2796E-AI camera and it works great.
SecurityLightAdded

rroller:
Thanks!

rroller:
Sorry I had to revert this change. See https://github.com/rroller/dahua/issues/410
Can you try to figure out a safer way to load this

alpauna:
So would it be safer to be more specific on the model IP8M-2796E so
`or self.model.startswith("IP8M-2796E")` 
since it appears to be impacting certain models of IP8M cameras hate to be so specific but I lost some functionality by the rollback.

PR #408, opened 2024-10-31
Update __init__.py

Pastaloverzzz:
Added support for CORE 2024.11

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -14,10 +14,11 @@ import hashlib
 
 from aiohttp import ClientError, ClientResponseError, ClientSession, TCPConnector
 from homeassistant.config_entries import ConfigEntry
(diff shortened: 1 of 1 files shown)


rroller:
I’m merging this without testing since I’m away from a computer for a bit. But it seems it won’t make things worse

Pastaloverzzz:
Redownloaded the integration again (version 0.9.63) and i'm not getting any errors, all entities are showing up and the button for adding new devices works as well so looks good to me.

Only thing that isn't correct is the version it displays in the integration, it still says 0.9.60. 

afbeelding

Thanks @rroller 🥇

rroller:
Thanks. Will fix the version later once I'm home at my computer. Amazing we can merge code from 30,000 feet in the air, but too hard to edit code on my mobile phone. Glad everyone is unblocked and thanks for the PR!

jodomi:
Redownloaded the integration (version 0.9.63) everything OK.

Thanks

PR #416, opened 2024-11-08
Re Added Security Light for more specific models IP8M-2796E 

alpauna:
Addressed the overly generic use of IP8M and made the security option more specific to IP8M-2796E since this model supports the security light feature. Addressed issue #405


rroller (approved):

rroller:
Thank you!

Issue #321, opened 2023-12-10
Support for LED illuminator on IPC-Color4K series (including IPC-Color4K-T)


brendanhaddock:
It is not currently possible to control the illuminator via the integration for my IPC-Color4K-T.

I would like to turn on the illuminator in response to motion. In my case - when someone walks into a specific zone configured in Frigate. 

Refer to [this post](https://ipcamtalk.com/threads/ipc-hdw5849h-ase-led-question-api-call-for-illuminator.69271/post-756113) linking to [this post](https://ipcamtalk.com/threads/worlds-first-review-dahua-dh-ipc-hdw5849h-ase-led-ipc-color4k-t-2-8mm-turret.64704/post-711029) on IPCamTalk. Essentially - they were unable to control the illuminator and instead had to resort to using a Day profile 24/7 and switching to a Night profile (with illuminator on) when required (e.g. motion occurred). 

I have recorded a network request via devtools. It successfully turns on the illuminator (with the broader camera profile set to "Self-adjustment"). See below. 

It doesn't work long term as it is session based, and I'm unsure how to convert it to a request format that uses auth parameters instead. 

Hoping that this project can add support, or at the very least someone can share a CURL request that we can use to call when we want to trigger the LED on/off. I've found many people after this functionality online, but no one who has cracked in.

Thank you to the devs of this project. 

**Functioning illuminator call (logged via DevTools)**
`curl 'http://192.168.5.92/RPC2' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: en-NZ,en-GB;q=0.9,en-US;q=0.8,en;q=0.7' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: username=admin; WebClientHttpSessionID=f080d6f18b287afbca3232e8ae25ce67' \
  -H 'Origin: http://192.168.5.92' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://192.168.5.92/' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36' \
[...]

PR #430, opened 2024-11-16
Illuminator support for IPC-HFW2439SP-SA-LED-S2

adobrotvorskiy:
Added check for Lighing_V2 support and API call if it supported in init.py

Also edit check for infrared light support. IPC-HFW2439SP-SA-LED-S2 has no infrared light but it was found by integration.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -131,6 +131,8 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         self._address = address
         self._max_streams = 3  # 1 main stream + 2 sub-streams by default
 
+        self._supports_lighting_v2 = False
+
         # channel_number is not the channel_index. channel_number is the index + 1.
         # So channel index 0 is channel number 1. Except for some older firmwares where channel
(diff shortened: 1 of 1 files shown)


rroller:
Thank you!

Pastaloverzzz:
After this update i got a depreciation warning. Not sure if this is the problem but since it seems to be something about the light.. 

I al Running 2024.12.b1 core but didn't get a error before i updated to the latest dahua version.

Logger: homeassistant.components.light
Bron: components/light/__init__.py:1329
integratie: Licht (documentatie, problemen)
Eerst voorgekomen: 22:59:08 (4 gebeurtenissen)
Laatst gelogd: 22:59:18

Entity None (<class 'custom_components.dahua.light.DahuaIlluminator'>) is using deprecated supported features values which will be removed in HA Core 2025.1. Instead it should use <LightEntityFeature: 0> and color modes, please create a bug report at https://github.com/rroller/dahua/issues and reference https://developers.home-assistant.io/blog/2023/12/28/support-feature-magic-numbers-deprecation

mrzv:
Is this related to, or can it help #321?

PR #437, opened 2024-12-17
Add support for Imou DB61i same as Amcrest AD410.

rwalker777:
Added support for the Imou DB61i which is identical to the Amcrest AD410.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -564,7 +564,7 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         IPC-HDW3849HP-AS-PV does https://dahuawiki.com/Template:NameConvention
         Addressed issue https://github.com/rroller/dahua/pull/405
         """
-        return "-AS-PV" in self.model or self.model == "AD410" or self.model.startswith("IP8M-2796E")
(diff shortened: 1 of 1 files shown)


rroller (approved):

rroller:
Thanks!

jrb80:
@rwalker777 I've been using the DB61i for a few years now with this integration, how does this commit change anything? Fyi, Imou is the consumer part of Dahau, and Dahua manufacture the AM410 for Amcrest (they are just a marketing company/brand).

rwalker777:
This gives doorbell press, ability to turn motion detect on and off and
control of the light.

jrb80:
The doorbell button has always worked with the `binary_sensor.doorbell_alarm_local` entity and it continues to work that way, and yes now a dedicated button entity is created `binary_sensor.doorbell_button_pressed `

This is the result that I get when pressing the button, is this the expected results? The `binary_sensor.doorbell_button_pressed ` state says `Detected sound` and `Clear`. Whilst the  `binary_sensor.doorbell_alarm_local` entity goes between `Safe` and `Unsafe` which remains unchanged. Perhaps the button entity state description is incorrect?

Btw, I appreciate the new functionality ;-)

Button Pressed

Alarm Local

rwalker777:
That is exactly how the AD410 always worked.  Changing it now would be breaking...

Issue #440, opened 2025-01-02
async_config_entry_first_refresh deprecated


rufik:
I've hit a message when pressed "Configure" button at integration page:
```
Detected that custom integration 'dahua' uses `async_config_entry_first_refresh`, which is only supported for coordinators with a config entry at custom_components/dahua/__init__.py, line 80: await coordinator.async_config_entry_first_refresh(). This will stop working in Home Assistant 2025.11, please create a bug report at https://github.com/rroller/dahua/issues
```

Not an error but worth to get fixed in next few months :)

rroller:
I can't reproduce this. What version of the plugin are you on? This should have been fixed already

rufik:
HA Core 2024.12.5
Dahua Integration 0.9.70
I'll try to reproduce it...

srkoster:
The deprecation message appears in the log when you trigger the OptionsFlowHandler (go to the integration and press `configure` next to the device).

I've drafted #447 to fix it

rroller:
Thanks! Released in https://github.com/rroller/dahua/releases/tag/0.9.71

eximius313:
I'm using `0.9.71` and the warning is still there (from `helpers/frame.py:324`):
```
Detected that custom integration 'dahua' uses `async_config_entry_first_refresh`, which is only supported for coordinators with a config entry at custom_components/dahua/__init__.py, line 80: await coordinator.async_config_entry_first_refresh(). This will stop working in Home Assistant 2025.11
```

srkoster:
@eximius313 that's strange because I'm running `0.9.71` and I'm not getting the error anymore.

When does it occur? And did you try removing the integration through HACS and doing a fresh install again through HACS?

eximius313:
> When does it occur?
I don't remember, I'm sorry :(

> And did you try removing the integration through HACS and doing a fresh install again through HACS?
it was fresh install, but when I reinstalled it now, I don't see it (maybe it occurs only with the first install)

zios0007:
This is fixed. The coordinator is constructed with its config entry:

```python
super().__init__(
    hass,
    _LOGGER,
    config_entry=entry,
    name=DOMAIN,
    update_interval=get_configured_scan_interval(entry),
)
```

which is precisely what the deprecation asked for: `async_config_entry_first_refresh` is supported for coordinators that have a config entry, and now it does.

[...]

PR #447, opened 2025-01-23
Removed manual setting of self.config_entry

srkoster:
This is a fix for #440 and removes the deprecated option of setting the self.config_entry in the OptionsFlowHandler (see https://developers.home-assistant.io/blog/2024/11/12/options-flow/)

Successfully tested locally


rroller:
Thank you!

PR #466, opened 2025-03-16
Added password and username url escaping to make it possible to use special chars in password

stalniy:
## Why

Currently integration doesn't support password with special characters because there is no proper escaping of characters

## What 

Added password and username url escaping to make it possible to use special chars in password


arturoparedesq:
Hi, I don't find where I can activate the siren to put auto, manual, off, I have a NVR4208-8P-AI

PR #501, opened 2025-10-03
Fix event stream 5-minute timeout disconnections

rmaher001:
## Why

The event stream disconnects every 5 minutes, causing the integration to miss important IVS events during reconnection.

## What

- Set `ClientTimeout(total=None)` on the event stream request to disable aiohttp's default 300-second timeout
- Allows the event stream to run indefinitely as intended
- Tested for 24+ hours without disconnections


rroller:
Thank you!

PR #508, opened 2025-10-05
Fix event stream timeout at session level

rmaher001:
## Summary
Moves timeout fix from request level (#501) to session level, which is more appropriate since the session is created specifically for the Dahua client.

## Changes
- Add `timeout=ClientTimeout(total=None)` to ClientSession creation in `__init__.py`
- Revert request-level timeout fix from `client.py`

## Rationale
The session-level fix prevents the default 5-minute timeout from disconnecting infinite event streams while keeping regular API requests properly scoped.

## Testing
Tested in production with no timeout errors.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -12,7 +12,7 @@ from datetime import timedelta
 from homeassistant.components.tag import async_scan_tag
 import hashlib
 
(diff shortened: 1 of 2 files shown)


rroller (approved):

rmaher001:
Thanks Ronnie for the merge and for making the Dahua cams in HA work.

- Richard

Issue #379, opened 2024-08-20
Privacy mode


akominch:
Hi, I have a DH-SD-H4C-0400B camera, it has a privacy mode where its lens goes down. When I enable this mode in DMSS, the following entry appears in the camera logs 
```
Name: leLensMask
State: Start
Result: Success
```
Maybe someone knows if there is any way to manage this locally via cgi or rpc?

Hukuma1:
Just switched from Amcrest to Dahua integration  and would also love the ability to flip on the Privacy Mode switch for some cameras that I have that support it (model IP4M-1041W if it matters). The official Amcrest integration has it, would it be possible to port it from there maybe?

beckynet:
I've DH-H4C This camera also has a Pivacy mode.

Hukuma1:
@zios0007 This is a longshot, and I apologize for being rude, but I saw you do a bunch of PRs to this repo. Any chance you could please add this? The code already exists in the official repo, perhaps we could merge it over to this repo? Again, sorry for rudely pinging you directly.

zios0007:
@Hukuma1 not rude at all, and you were closer than you knew — the code you are
asking for has existed in this repo for eleven months.

**#516, "Add Privacy Mode Toggle" by @AdrianCzebiniak**, opened 2025-10-12. It
adds a Privacy Mode switch driven by `LeLensMask` over RPC2 — which matches the
`Name: leLensMask` line @akominch found in their camera log at the top of this
issue. It is 301 lines across 10 files with its own tests.

It never stalled on review. @rroller left one comment, the author fixed it two
days later, and then `main` moved underneath it until it went `CONFLICTING`.
That is the whole story: it needed a rebase, not a decision.

So I have done that — **#682** carries @AdrianCzebiniak's commit, authorship
intact, rebased onto current `main`. The three conflicts were all against
changes of mine and all additive. If it merges, this issue closes with it.

Two honest caveats:

- I have no camera with `LeLensMask`, so this is unverified on hardware — as
  #516 was. You have an IP4M-1041W and @beckynet has a DH-H4C; once it is
  released, a "the lens actually drops" from either of you is the confirmation
  that is missing.
- It is @rroller's call whether to take it, and I have flagged one thing in the
  PR for him to look at — the RPC2 session is opened and closed per poll, which
  cuts against a lot of recent work reducing logins.

@beckynet @GGontijo @akominch — tagging you since you all asked here.

[...]

PR #516, opened 2025-10-12
Add Privacy Mode Toggle

AdrianCzebiniak:
Added a privacy mode toggle.  Tested on an Amcrest IP4M-1041


rroller:
has merge conflicts

zios0007:
@AdrianCzebiniak I have rebased this onto current `main` and opened it as #682,
with your commit and your authorship intact. Apologies for doing that rather
than asking first — #379 has been open since August 2024 and someone asked again
yesterday, so I went for the thing that moved it.

**Nothing was wrong with your code.** You answered @rroller's review point two
days after he made it; what stalled this was eleven months of `main` moving
underneath it. All three conflicts were in `__init__.py` and all three were
against changes of mine, all additive: the `IPC-COLOR4M-TZ` LightingScheme probe,
its per-poll read, and `get_vto_client` gaining a `| None` return. Each resolved
by keeping both sides. Same 10 files, same +301/-7.

I added one commit of my own: the new switch needed an entry in
`tests/dahua/test_entity_identity.py`, which is the ledger of every unique_id the
integration produces. One of its tests asserts the whole set as a literal, so it
would have failed without it.

I also flagged one thing for review rather than changing it — `_async_privacy_mode_rpc2`
logs in and out per call, so a supporting camera polls a login every 30 seconds
by default, which cuts against #616/#628/#632/#639. It may be deliberate, and it
is your design decision, not mine to quietly rewrite.

**If you would rather rebase this yourself, say so and I will close #682.** That
is genuinely the outcome I would prefer.

zios0007:
@AdrianCzebiniak this is now in — #682 merged and went out in **0.10.2**, with
your commit and your authorship on it. #379 closed with it, after two years open.

This PR can be closed as superseded whenever you or @rroller get to it; I cannot
close it myself. Nothing of yours was dropped — same 10 files, same +301/-7, the
only addition being the unique_id ledger entry the new switch needed.

Thank you for writing it. It sat eleven months for want of a rebase, which is a
poor return on work that was finished and correct.

One thing still worth your eye, which I flagged in #682 rather than changing:
`_async_privacy_mode_rpc2` logs in and out per call, so a supporting camera polls
a login every 30 seconds at the default interval. If that was deliberate, it is
worth a comment in the code saying so; if not, `_shared_rpc2()` already holds one
session per host and would be a small follow-up.

PR #542, opened 2026-02-10
Fix session leaks, replace blocking threads with async tasks, add reauth

brianegge:
## Summary
- **Fix session leaks**: Close `aiohttp.ClientSession` in config flow credential test (`finally: await session.close()`) and on integration unload (`await coordinator.async_stop()`)
- **Replace blocking threads with async tasks**: Remove `DahuaEventThread` and `DahuaVtoEventThread` (and `thread.py` entirely), replacing them with native `asyncio.Task` coroutines that run on HA's event loop. VTO client now uses `loop.call_later()` instead of `threading.Timer` and a `disconnected` future instead of `loop.stop()`
- **Fix swallowed exceptions**: `client.stream_events()` no longer silently catches `CancelledError` or swallows all exceptions with bare `except: pass`
- **Add reauthentication flow**: Detects HTTP 401 during initialization and triggers HA's reauth UI so users can update credentials without removing/re-adding the integration

## Test plan
- [x] Deploy to live HA instance with multiple Dahua cameras
- [x] Restart HA and verify all cameras initialize without errors
- [x] Verify camera streams still work in dashboard
- [x] Verify motion/event detection still fires `dahua_event_received` events
- [x] Test reauth flow by temporarily changing camera password


basp89:
Hi Brian, rroller, thanks for updating. Unfortunately after updating  to anything from 0.9.76 up, scanning an rfid tag crashes home assistant. Anything you could do about that perhaps?

brianegge:
It's related to this change? Is there a stack trace or issue open?

PR #561, opened 2026-03-23
Fix: return list from translate_event_code to dispatch all matching listeners

### LINKED ISSUE #560: CrossLine not working any more after 0.9.8 Fix
After patch "Fix SmartMotionHuman/Vehicle not triggered with CrossLine/CrossRegion events
#557" CrossLine do not work any more. I've got automations with Car CrossLine and it did not work tonight when the car crossed the line. After that I have checked the activity in Integration and it not triggered either. But Dahua's Interface have detected it clearly

MicheleMercuri:
## Summary

Fixes #560 — regression from #557 where `CrossLineDetection` sensors stopped firing when `SmartMotionHuman`/`SmartMotionVehicle` listeners also existed.

**Root cause:** `translate_event_code()` returned a single event code string. When both a CrossLine and a SmartMotion listener were registered, the CrossLine event was exclusively translated to SmartMotion — the original `CrossLineDetection` sensor never fired.

**Fix:** `translate_event_code()` now returns a `list[str]` of event codes, and both callers (`on_receive_vto_event`, `on_receive`) iterate over the list to dispatch all matching listeners.

### Behavior matrix

| User config | CrossLine + ObjectType=Human | Result |
|---|---|---|
| Only CrossLine subscribed | `[CrossLineDetection]` | works (unchanged) |
| Only SmartMotion subscribed | `[SmartMotionHuman]` | works (unchanged) |
| Both subscribed | `[CrossLineDetection, SmartMotionHuman]` | **both fire** (was broken) |
| Neither subscribed | `[SmartMotionHuman]` fallback | works (unchanged) |

### Changes (3 touch points in `__init__.py`)

1. **`translate_event_code()`** — returns `list[str]` instead of `str`. Builds a list: always includes original CrossLine/CrossRegion if its listener exists, then appends SmartMotion* if that listener also exists.
2. **`on_receive_vto_event()`** — `codes = ...` + `for code in codes:` loop around the existing handler body.
3. **`on_receive()`** — `event_names = ...` + `for event_name in event_names:` loop around the existing handler body.

## Test plan

- [x] Verified on 4 Dahua cameras with IVS CrossLineDetection + SmartMotionHuman both enabled — both `binary_sensor.*_crosslinedetection` and `binary_sensor.*_smart_motion_human` fire on the same event
- [x] Verified CrossLineDetection-only setup (no SmartMotion entities) still works
- [x] Verified SmartMotionHuman-only setup (no CrossLine entities) still works
- [x] Verified doorbell (BackKeyLight → DoorbellPressed) still works (returns `["DoorbellPressed"]`)
- [x] Verified VTO events (AccessControl, DoorStatus Pulse) still work

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -477,40 +477,42 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         # }
 
         # This is the event code, example: VideoMotion, CrossLineDetection, BackKeyLight, PhoneCallDetect, DoorStatus, etc
-        code = self.translate_event_code(event)
-        event_key = self.get_event_key(code)
+        codes = self.translate_event_code(event)
 
-        if code == "AccessControl":
-            card_id = event.get("Data", {}).get("CardNo", "")
-            if card_id:
-                card_id_md5 = hashlib.md5(card_id.encode()).hexdigest()
-                self.hass.async_create_task(
-                    async_scan_tag(self.hass, card_id_md5, self.get_device_name())
-                )
+        for code in codes:
+            event_key = self.get_event_key(code)
 
-        listener = self._dahua_event_listeners.get(event_key)
-        if listener is not None:
-            action = event.get("Action", "")
-            if action == "Start":
-                self._dahua_event_timestamp[event_key] = int(time.time())
-                listener()
-            elif action == "Stop":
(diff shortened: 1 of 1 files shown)


rroller (approved):

rroller:
Thanks!

PR #562, opened 2026-03-29
Fix disarming linkage 400 error on cameras without channel-indexed config

irfadrazick:
## Summary
- Some cameras (e.g. DH-P3D-3F-PV-P) return `400 Bad Request` when setting `DisableLinkage` and `DisableEventNotify` with a channel index, even though the read (`getConfig`) succeeds
- Fall back to the non-indexed URL if the channel-indexed `setConfig` call fails

## Test plan
- Tested on DH-P3D-3F-PV-P (firmware 3.000.0000000.6.R, build:2025-09-25)
- Toggling Disarming and Event Notifications switches in Home Assistant now works without error

### DIFF
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -630,7 +630,12 @@ class DahuaClient:
             value = "true"
 
         url = "/cgi-bin/configManager.cgi?action=setConfig&DisableLinkage[{0}].Enable={1}".format(channel, value)
-        return await self.get(url)
+        try:
+            return await self.get(url)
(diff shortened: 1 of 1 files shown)


irfadrazick:
Hey @rroller, I tested this change with my camera by modifying the code directly in my HA and actively using it. It would be great if you could check and merge it so I can install the latest updates of the integration when you release.

rroller:
Thanks! Will get this merged soon. Sorry been busy

rroller (approved):

rroller:
Released!

Issue #477, opened 2025-04-12
FYI: added DH-P5D-5F-PV - working in general


Masterz69:
Recently added dual camera device DH-P5D-5F-PV.
Integration added to HA "PT Lens" camera streams, but not "Fixed Lens".
As well following message appears in log, probably related to streams count.

Let me know if somehow can help to looks into details.

```
This error originated from a custom integration.

Logger: custom_components.dahua
Source: custom_components/dahua/client.py:745
integration: Dahua (documentation, issues)
First occurred: 9 April 2025 at 07:34:08 (5 occurrences)
Last logged: 11 April 2025 at 18:12:38

too many values to unpack (expected 2)
Traceback (most recent call last):
  File "/config/custom_components/dahua/client.py", line 745, in stream_events
    on_receive(data, channel)
    ~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/config/custom_components/dahua/__init__.py", line 457, in on_receive
    events = parse_event(data)
  File "/config/custom_components/dahua/dahua_utils.py", line 71, in parse_event
    key, value = key_value.split('=')
    ^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

Masterz69:
During last night HA lost connection to camera - all entities unavailable.
After Integration reload - it show 2 devices.
* With model DH-P5D-5F-PV - added manually and now unavailable;
* Generic RSTP - added on reload, entities available, 4x streams added, but no picture/stream.

![Image](https://github.com/user-attachments/assets/d90be226-3a5c-4358-a369-c745c2f167b3)

![Image](https://github.com/user-attachments/assets/ece74ce8-2113-4a4f-a9fe-80d687c8a9aa)

In Log:
```
Logger: homeassistant.components.stream.stream.camera.ipc_back_sub_2_2
Source: components/stream/__init__.py:489
integration: Stream (documentation, issues)
First occurred: 08:45:03 (7 occurrences)
Last logged: 08:48:34

Error from stream worker: Forbidden error opening stream (HTTP_FORBIDDEN, Server returned 403 Forbidden (access denied), rtsp://****:****@192.168.1.92:554/cam/realmonitor?channel=1&subtype=2)
```
probably because Generic RSTP ctreated with no valid credentials provided ?

inhibitz:
I also have the identical problem.

AhmadMuj:
I have this problem as well

piecyk57:
[...]

PR #563, opened 2026-04-02
fix: use split('=', 1) in parse_event to handle '=' in values (#477)

jcass77:
Fix for #477: parse_event crashes with ValueError when event data contains '=' characters (e.g. base64-encoded values). The split('=') call produces more than 2 parts, causing 'too many values to unpack'.

Fix: use split('=', 1) to split on the first '=' only, preserving any '=' characters in the value portion.

Adds unit tests for parse_event covering the bug and happy paths.

### DIFF
--- a/.agents/summary/architecture.md
+++ b/.agents/summary/architecture.md
@@ -1,0 +1,99 @@
+# Architecture
+
+## System Architecture
+
+```mermaid
+graph TB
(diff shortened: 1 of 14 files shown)


rroller (approved):

rroller:
Thanks!

PR #579, opened 2026-05-24
Add option to disable channel-number auto-detection

### LINKED ISSUE #578: RTSP fails with channel=0 when camera also responds to HTTP snapshot at index 0
# RTSP fails with `channel=0` when camera also responds to HTTP snapshot at index 0

## Summary

For some Dahua cameras, the channel auto-detection heuristic in `_async_update_data` (`__init__.py` ~line 268-276) incorrectly resets `_channel_number` to `0` because the camera responds to `/cgi-bin/snapshot.cgi?channel=0`. The resulting RTSP URL uses `channel=0`, and the camera then returns HTTP 404 on the RTSP request — the stream is broken indefinitely until the camera reboots into a different state.

## Environment

- Integration version: **0.9.83** (HACS, latest)
- Home Assistant: 2026.5.x
- Camera model reported via Dahua API: `IPC-...` (serial `4E03EC9PAGC9647`, firmware date is recent post-power-loss reboot)
- Three other Dahua cams on the same install (same brand, similar firmware era) are unaffected — their snapshot(0) fails, the heuristic correctly leaves `_channel_number = channel + 1 = 1`, RTSP works.

## What happens

1. Camera comes back up after a power outage.
2. Integration's coordinator runs `_async_update_data`. It calls `async_get_snapshot(0)`. The camera **returns a valid JPEG** for that request (HTTP CGI endpoint is happy with channel index 0).
3. The "channel reset" branch fires:
   ```python
   if not self.is_doorbell():
       self._channel_number = self._channel  # == 0
   ```
4. From that point the camera entity is initialized with `_stream_source = rtsp://...:554/cam/realmonitor?channel=0&subtype=0`.
5. Every stream attempt logs:
   ```
   ERROR (stream_worker) [homeassistant.components.stream.stream.camera.camera_X_main]
   Error from stream worker: Not Found error opening stream
   (Server returned 404 Not Found, rtsp://****:****@camera-X:554/cam/realmonitor?channel=0&subtype=0)
   ```
6. The entity stays usable for snapshots (HTTP at channel 0 still works) but every recorder/preview/automation that depends on RTSP fails. In our case, `camera.record` calls hung an `automation: single` notifier in "Already running" for 12+ hours.

Manually probing the camera confirmed: HTTP snapshot works on both 0 and 1; RTSP only works on `channel=1`.

## Reproduction

I can repro by manually setting `data.channel = 0` in `.storage/core.config_entries` for the affected camera and restarting HA. Forcing `data.channel = 1` doesn't help on its own because the auto-detect still resets it during refresh (and additionally creates a duplicate set of `*_2` entities because the unique_id-derived suffix changes).

## What I'd like

A way for users to disable the channel-reset heuristic on a per-camera basis. The current behavior is correct for cameras the heuristic was written for, but there's no escape hatch when it gets it wrong on a given firmware.

Happy to send a PR — proposed approach is a new `auto_detect_channel: bool` option (default `True`, so behavior is unchanged for everyone) surfaced in the OptionsFlow alongside the existing platform toggles. When the user flips it off, the snapshot probe and reset are skipped.

## Workaround currently in production

Local patch that comments out the snapshot reset block. Works, but fragile across integration updates.

YusDyr:
# Add option to disable channel-number auto-detection

Closes #578.

## Problem

The coordinator's `_async_update_data` probes `async_get_snapshot(0)` and, on success, resets `self._channel_number = self._channel` (line ~272 in `__init__.py`). The heuristic exists because some Dahua firmwares index channels from 0 instead of 1, and the camera's HTTP CGI endpoint reveals which family it belongs to.

On some cameras (full details in the linked issue) the HTTP snapshot endpoint cheerfully returns a JPEG at index 0 even though RTSP only streams on `channel=1`. The heuristic then mis-classifies the camera as 0-indexed and the resulting `rtsp://.../cam/realmonitor?channel=0&subtype=0` returns HTTP 404 forever after.

There's currently no way for affected users to opt out.

## Change

Add an `auto_detect_channel` option to the integration's OptionsFlow (default `True`, identical to current behavior). When disabled, the snapshot probe and channel reset are skipped, and `_channel_number` stays at `channel + 1`.

Existing installs are not affected — the new key isn't present in their `options` dict, so `options.get(CONF_AUTO_DETECT_CHANNEL, True)` returns the default `True` and the original code path runs.

## Files

- `const.py` — new `CONF_AUTO_DETECT_CHANNEL` constant.
- `__init__.py` — gate the snapshot reset on the new option, mention the toggle in the DEBUG log.
- `config_flow.py` — surface the toggle in the existing `DahuaOptionsFlowHandler` schema alongside the platform booleans.

## How to use

After updating, **Settings → Devices & services → Dahua (the affected camera) → Configure → uncheck "auto_detect_channel"**. The change takes effect on the next coordinator refresh (or immediately if the entry is reloaded).

## Testing

Verified on a real install: with the toggle off, `_channel_number` stays at 1, RTSP stream comes up, `camera.record` produces a valid MP4. Other cameras in the same install (which rely on the heuristic) keep the option on and are unaffected. `python -c "import ast; ast.parse(open(f).read())"` clean on the three changed files.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -36,6 +36,7 @@ from .const import (
     CONF_RTSP_PORT,
     STARTUP_MESSAGE,
     CONF_CHANNEL,
+    CONF_AUTO_DETECT_CHANNEL,
 )
 from .dahua_utils import parse_event
 from .vto import DahuaVTOClient
@@ -265,15 +266,22 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                 self.machine_name = data.get("table.General.MachineName")
                 self._serial_number = data.get("serialNumber")
 
-                try:
-                    await self.client.async_get_snapshot(0)
-                    # If able to take a snapshot with index 0 then most likely this cams channel needs to be reset
-                    # but check if unit is not a doorbell first as channel 0 doesnt exist for VTOs
-                    if not self.is_doorbell():
-                        self._channel_number = self._channel
-                except ClientError:
-                    pass
-                _LOGGER.debug("Using channel number %s", self._channel_number)
+                # Some Dahua firmwares index channels from 0, others from 1. The default
+                # is to auto-detect: if a snapshot at index 0 succeeds, treat this camera as
+                # 0-indexed and reset channel_number accordingly. Users on cameras where this
+                # heuristic gets it wrong (HTTP snapshot at 0 succeeds but RTSP only streams
+                # on channel=1) can disable it via the integration options.
+                auto_detect = self.config_entry.options.get(CONF_AUTO_DETECT_CHANNEL, True)
+                if auto_detect:
+                    try:
+                        await self.client.async_get_snapshot(0)
+                        # If able to take a snapshot with index 0 then most likely this cams channel needs to be reset
+                        # but check if unit is not a doorbell first as channel 0 doesnt exist for VTOs
+                        if not self.is_doorbell():
+                            self._channel_number = self._channel
+                    except ClientError:
+                        pass
+                _LOGGER.debug("Using channel number %s (auto_detect=%s)", self._channel_number, auto_detect)
 
                 try:
                     await self.client.async_get_coaxial_control_io_status()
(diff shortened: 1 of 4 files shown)


YusDyr:
N.B. Prepared with Claude Code usage (opus 4.7 model).

rroller (approved):

rroller:
thanks!

PR #581, opened 2026-06-01
Fixed illuminator state/control on dual illuminator camera (IL series)

Trexano99:
✨ Fix: Dynamic illuminator indexing for Dahua dual smart light cameras (IL Series)

### 🚨 The Problem
On modern Dahua dual smart light cameras (IL series), the illuminator state and functionality were bugged. Previously, the white light was implicitly assumed to be at `index 0`. While this works for single-illuminator cameras, on dual-illuminator models, `index 0` is assigned to the **Infrared (IR) light**, which breaks white light controls.

### 🛠️ The Solution
* **Introduced `get_illuminator_index`:** A new function to dynamically identify the correct light index regardless of the camera's hardware layout, cleanly preserving legacy behavior.
* **Refactored `async_set_lighting_v2`:** Parameterized the index within the function call, ensuring the dynamically identified light index is resolved *before* execution.

### ✅ Impact
* 💡 Successfully restores **turn on/off** and **brightness** controls for the white light on IL series cameras.
* 🛡️ Ensures strict backward compatibility for single-illuminator models.

### 📝 Documentation & Usage Note
To control the illuminator via the Home Assistant entity, the camera's illuminator mode **must** be set to `manual`. 
*(Tip: Setting the night profile to "manual illuminator" allows for consistent manual control directly from the HA entity).*

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -362,6 +362,15 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                     except ClientError:
                         _LOGGER.debug("Cam does not support profile mode. Will use mode 0")
                         self._supports_profile_mode = False
+                    if not self._supports_profile_mode:
+                        # The legacy Lighting[0][2] probe is empty on some cams (e.g. white-light/dual
+                        # illuminator models). Fall back to the VideoInMode API, which is the canonical
+                        # source for the active day/night profile.
+                        try:
+                            mode_data = await self.client.async_get_video_in_mode()
+                            self._supports_profile_mode = "table.VideoInMode[0].Config[0]" in mode_data
+                        except ClientError:
+                            self._supports_profile_mode = False
                     _LOGGER.debug("Device supports profile mode=%s", self._supports_profile_mode)
                 else:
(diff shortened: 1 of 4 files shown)


rroller:
has merge conflicts

zios0007:
@Trexano99 your diagnosis here was right, and the problem is fixed, though not by this PR.

Index 0 being the infrared emitter on dual light models, with the white light at 1, is exactly what was happening. It landed in #652 (released 0.9.95) as `get_illuminator_index`, resolved from the device's own `LightType` rather than assumed. Two follow-ons went with it: #659 (0.9.96) for the brightness bank, since some models put the white light's level on `NearLight` rather than `MiddleLight`, and #654 (0.9.97) which warns when `LightingScheme` is in a mode that will not let the white light come on whatever is written.

So there is nothing left for this branch to carry, and it has been conflicting since August. Worth closing unless you can see something in it that the above missed.

Flagging one thing in case it is not obvious, because it applies to the other stale PRs here too: while a PR conflicts, GitHub cannot build the test-merge commit, so it never starts a CI run at all. It is not queued or slow, it is simply never dispatched. That is easy to read as nobody looking.

PR #582, opened 2026-06-01
Fixed day/night profile control on dual illuminator cameras (IL series)

Trexano99:
✨ Fix: Reliable day/night profile switching for Dahua dual smart light cameras (IL Series)

🚨 The Problem
On modern Dahua dual smart light cameras (IL series), dahua.set_video_profile_mode didn't hold. Forcing Day (or Night) appeared to work for a moment, then the camera would flip back on its own and keep oscillating between day and night indefinitely.

There were two root causes:

The service wrote VideoInMode[0].Config[0], which these cameras ignore,  they select the active profile through the ConfigEx string, not the Config index list (which stays a static [0,1]).
Even once the profile was set, the camera was left on its automatic switch mode (Mode=2, by brightness), which reverts any forced profile within seconds. 
As a side effect, the integration read the active profile from Config[0] (always 0 on these cams), so the illuminator entity always tracked the day profile.

🛠️ The Solution
Introduced async_set_video_profile_config_ex: forces the profile exactly the way the camera's own web UI does: VideoInMode.Mode=4 (manual) + VideoInMode.ConfigEx=Day/Night, so the requested profile actually sticks.
Added ConfigEx auto-detection (supports_config_ex): cameras that report ConfigEx use the new path; everything else cleanly preserves the legacy Config[0] behavior.
Fixed the read side: _profile_mode is now derived from ConfigEx on these cameras, so the illuminator state/brightness follow the correct active profile.

✅ Impact
💡 set_video_profile_mode now reliably forces Day/Night on IL series cameras with no oscillation.
🛡️ Strict backward compatibility for cameras that use the legacy Config[0] path.
🔬 Verified against a live device: Mode=2 + forced Night reverts to Day in ~8s, while Mode=4 + forced Night holds indefinitely.

📝 Documentation & Usage Note
Because the fix writes a persistent config, after calling set_video_profile_mode the camera stays in manual switch mode (Mode=4), automatic day/night switching remains off until you re-enable it from the camera's web UI. This is intended for forcing a profile from HA (e.g. to drive the illuminator) and mirrors what the camera UI's own manual toggle does.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -362,6 +362,15 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                     except ClientError:
                         _LOGGER.debug("Cam does not support profile mode. Will use mode 0")
                         self._supports_profile_mode = False
+                    if not self._supports_profile_mode:
+                        # The legacy Lighting[0][2] probe is empty on some cams (e.g. white-light/dual
+                        # illuminator models). Fall back to the VideoInMode API, which is the canonical
+                        # source for the active day/night profile.
+                        try:
+                            mode_data = await self.client.async_get_video_in_mode()
+                            self._supports_profile_mode = "table.VideoInMode[0].Config[0]" in mode_data
+                        except ClientError:
+                            self._supports_profile_mode = False
                     _LOGGER.debug("Device supports profile mode=%s", self._supports_profile_mode)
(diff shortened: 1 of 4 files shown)


rroller (approved):

rroller:
Sorry I had to revert this. See https://github.com/rroller/dahua/issues/605

PR #584, opened 2026-06-11
Use current profile to get brightness, default to 100

tylerransdell:
This does 2 things:
1. Uses current profile to get current brightness instead of pulling from profile 0. 
2. If no brightness found, returns 100 instead of 100/255*100 = 39, discard if this was intentional. I believe this approach is much more automation friendly.

 Disclaimer:
1. cline shell with model openrouter/owl_alpha (AI) help diagnose and repair issues. 
2. I fully understand the changes I made.
3. This shouldn't change any other behaviors.
4. I do not fully understand the entire codebase why the old code was the way it was.

Testing:
This works for white light cameras N45EJ62 (IPC-HDW5442TP-S-LED)


tylerransdell:
More changes, new descriptions needed.

PR #587, opened 2026-07-24
Fix hybrid/thermal cameras misdetected as doorbells, dropping IVS events

jaaneo:
## Problem

I have a DHI-TPC-BF1241 (hybrid visual+thermal camera) alongside several regular IPC-HFW3849T1-AS-PV cameras. The regular cameras receive line-crossing/region events fine, but the hybrid camera never received any events at all, not even `VideoMotion`.

## Root cause 1: `is_doorbell()` false positive

```python
def is_doorbell(self) -> bool:
    m = self.model.upper()
    return m.startswith("VTO") or m.startswith("DH-VTO") or (
        "NVR" not in m and m.startswith("DHI")) or ...
```

Any model starting with `DHI-` (and not containing `NVR`) is treated as a doorbell (VTO). That's too broad - it's meant to catch doorbell models like `DHI-VTO2211G-P`, but it also matches regular cameras that happen to use the common `DHI-` prefix, like `DHI-TPC-BF1241`. Once misclassified, the integration opens a VTO connection on port 5000 instead of the normal event stream, which the camera doesn't support, so it silently never receives any events.

Fix: match on the `VTO` substring instead, which still catches `VTO...`, `DH-VTO...`, `DHI-VTO...`, without matching unrelated `DHI-` prefixed models.

## Root cause 2: channel filter drops IVS events on hybrid cameras

After fixing `is_doorbell()`, the camera started connecting normally, but cross-line/cross-region events were still silently dropped:

```python
if index != self._channel:
    continue
```

This filter exists so NVRs/DVRs - where multiple config entries (one per channel) share one event stream - can discard events meant for other channels. But the DHI-TPC-BF1241 reports `CrossLineDetection`/`CrossRegionDetection` on an internal sub-channel index (`index: 1`) that differs from its configured video channel (`channel: 0`), so this filter dropped them too, even for a single standalone entry with no channel-sharing going on.

Fix: only apply the filter when it's actually needed - i.e. when more than one config entry points at the same device address. A standalone camera (including a hybrid one with only one entry) isn't sharing its event stream with anything else, so there's nothing to filter out.

## Update: further event loss found once split into two entries (visual + thermal)

Running this in production against a two-entry setup (visual on one channel, thermal on the other, both against the same device address) surfaced three more ways events were getting lost:

**Root cause 3: a second connection to the same device silently goes stale.** Once two config entries share a device address, each opened its own `attach` event-stream connection. The camera only pushes real event payloads to the *first* open session - a second, independent connection doesn't error, it just keeps receiving heartbeats and nothing else. Fix: one connection is now shared per physical device address; the coordinator that opens it fans every event out to all peer coordinators for that address (each peer still filters by its own channel per Root cause 2's fix), and ownership hands off to a peer if the owner disconnects.

**Root cause 4: TCP chunk boundaries don't line up with event boundaries.** Chunks from `response.content.iter_chunks()` were passed straight to `on_receive`, but the camera's multipart stream is delimited by `--myboundary`, not by TCP chunk boundaries - so a single event's JSON payload frequently got split across two chunks, either silently failing to parse or throwing mid-parse and killing the whole connection (dropping any paired Stop event behind it, leaving a `binary_sensor` stuck on). Fix: buffer bytes and only dispatch blocks bounded by two `--myboundary` markers.

**Root cause 5: `parse_event()` assumed a fixed number of header lines before `Code=`.** Events from the thermal channel don't always have the same header shape as the visual channel, so the fixed-offset split silently dropped them. Fix: search for `Code=` wherever it falls in the block instead of assuming its position.

Also added: exception isolation per individual event (so one malformed/unexpected payload doesn't take down the whole stream connection instead of just being skipped), and a `sock_read` timeout against the 5s heartbeat to detect a dead-but-silent connection (no FIN/RST) so it reconnects instead of hanging forever.

## Testing

- IPC-HFW3849T1-AS-PV (regular camera, unaffected by any of these bugs): confirmed still works and channel filtering still applies correctly.
- DHI-TPC-BF1241 added as a single entry: confirmed `is_doorbell()` now returns `False`, event stream connects normally, and `CrossRegionDetection`/`CrossLineDetection` events arrive in the debug log with `index: 1`.
- DHI-TPC-BF1241 split into two entries (visual on one channel, thermal on the other): confirmed via recorder history that a cross-region event correctly flips only the correct channel's sensor, and that both channels keep receiving live events over an extended run (no more silent stalls on the second entry, no truncated/dropped events at TCP chunk boundaries).

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -557,9 +557,12 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                 except ValueError:
                     index = 0
 
-            # This is a short term fix. Right now for NVRs this integration creates a thread per channel to listen to events. Every thread gets the same response. We need to
+            # This is a short term fix. Right now for NVRs/DVRs this integration creates a thread per channel to listen to events. Every thread gets the same response. We need to
             # discard events not for this channel. Longer term work should create only a single thread per channel.
-            if index != self._channel:
+            # Single cameras (including hybrid/thermal cameras that report events on an internal
+            # sub-channel index different from their configured channel) aren't sharing this stream
+            # with other config entries, so there's nothing to filter out.
+            if index != self._channel and self.is_multichannel_device():
                 continue
 
             # Put the vent on the HA event bus
@@ -657,8 +660,18 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
     def is_doorbell(self) -> bool:
         """ Returns true if this is a doorbell (VTO) """
         m = self.model.upper()
-        return m.startswith("VTO") or m.startswith("DH-VTO") or (
-            "NVR" not in m and m.startswith("DHI")) or self.is_amcrest_doorbell() or self.is_empiretech_doorbell() or self.is_avaloidgoliath_doorbell()
+        return "VTO" in m or self.is_amcrest_doorbell() or self.is_empiretech_doorbell() or self.is_avaloidgoliath_doorbell()
+
+    def is_multichannel_device(self) -> bool:
+        """
+        Returns true if another config entry shares this device's address (e.g. an NVR/DVR with one
+        entry per channel, or a hybrid camera added twice for its visual and thermal channels). In
+        that case every entry's event stream carries events for every channel, so each entry must
+        filter out events meant for the others. A device with only one config entry pointed at it
+        isn't sharing its stream, so nothing needs to be filtered out by channel index.
+        """
+        entries = self.hass.config_entries.async_entries(DOMAIN)
+        return sum(1 for e in entries if e.data.get(CONF_ADDRESS) == self._address) > 1
 
     def is_amcrest_doorbell(self) -> bool:
         """ Returns true if this is an Amcrest doorbell - IMOU DB61i is identical """
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -15,7 +15,7 @@ import hashlib
 from aiohttp import ClientError, ClientResponseError, ClientSession, TCPConnector
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import CALLBACK_TYPE, HomeAssistant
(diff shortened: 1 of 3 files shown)


rroller:
Can you fix the conflicts

zios0007:
@jaaneo I went through this properly today. Your diagnosis was good and most of it is now fixed, though piecemeal and by other PRs rather than by this branch, which is a poor outcome for the person who did the work.

| | root cause | state |
|---|---|---|
| 1 | `is_doorbell()` matched any `DHI-` model | fixed in #668 |
| 2 | channel filter drops sub-channel events | **still open** |
| 3 | a second connection to one device goes stale | fixed in #615 |
| 4 | TCP chunks split event boundaries | fixed in #678 and #709 |
| 5 | fixed header offset before `Code=` | #734, opened today |

Root cause 5 was still live and is your find. `parse_event` skipped three lines and needed the fourth to start with `Code=`, so a block with one header line fewer went into the `len(s) < 4` guard and was dropped. Every existing fixture uses the four-line shape, so the whole suite passes with the bug reintroduced. You are credited in the commit.

**Root cause 2 is the one still open**, and I have not touched it because I do not think the proposed fix is safe as written. Skipping the filter when only one config entry points at a device is right for a standalone camera. On a recorder with a single channel configured it is not: `_by_channel` then holds one coordinator, and every other channel's motion would land on that one entity. `is_nvr_channel()` already exists and would catch most of it, since a configured channel above 0 or a model saying NVR both answer yes, but it leans on the same "is this a recorder" question that has never had a clean answer here.

If you still have the DHI-TPC-BF1241, that is the missing piece. What does it report for `index` on `CrossLineDetection`, and what channel is the entry configured as? With that I would write it gated on `is_nvr_channel()` and say plainly what the residual gap is.

Also worth knowing, because it explains the silence: while a PR conflicts, GitHub cannot build the test-merge commit, so it never creates a CI run. Not queued, not slow, never dispatched. This has been in that state since August.

Issue #577, opened 2026-05-21
Dahua Integration has massive login/logout to Dahua NVR showing in Dahua NVR log


dawuhe:
In 24 hours time frame, there are total 114300 login and logout entries in NVR log for admin user and HA IP address that was used to add Dahua camera in Dahua integration in HA. All 5 cameras are using the same NVR "admin/password" and NVR IP, only difference is channel number. among these login/logout entry, it is about 50/50 for login/logout. every seconds there is (114300 / 24 hours /3600 seconds = ) 1.32 times login/logout activity. 

In HA Dahua integration, each camera only have 3 total entities while the rest of entities are disabled. These 3 entities are camera main stream, camera sub stream and binary sensor region cross.

Please investigate and resolve this issue.

Below are version and device details:

1. Dahua HA Integration: Version 0.9.81

2. Dahua device:
NVR: DHI-NVR4108-8P-4KS2
5 x POE Camera: IPC-HDBW4631R-ZS

JimGermany:
This problem has existed for years. :-D 

I haven't seen or received an explanation or even a solution for this anywhere in all these years. Therefore, I've long since integrated my Dahua NVR into HA via ONVIF. While this means I have to forgo a few features of the Dahua integration, at least it works flawlessly with ONVIF.

zios0007:
The mechanism behind this is that `DahuaClient` builds a fresh `DigestAuth` for every request:

```python
auth = DigestAuth(self._username, self._password, self._session)
response = await auth.request("GET", url)
```

A new `DigestAuth` has no challenge, so every single call starts unauthenticated, takes a `401`, then repeats the request with credentials. To the NVR that is a login and a logout per request — which matches the roughly 50/50 split of login to logout entries reported here, and the rate scaling with poll frequency rather than with the number of entities.

#597 holds the challenge on the client and shares it across requests, so a call only re-authenticates when the nonce is actually stale. That should cut the NVR log volume substantially rather than eliminating it — periodic re-auth on nonce expiry is still expected and correct.

For what it is worth, on my own NVR-backed setup (33 camera entities) I measured **3,358 unavailable/unknown transitions per day** across the camera entities before the change, which is 51% of all entity churn on that Home Assistant instance. Same shape as this report: many channels behind one NVR, one credential.

Worth testing against this issue once #597 lands, since I do not have a way to read NVR-side logs to confirm the count directly.

JimGermany:
Since I had posted something about this here as well:

[...]

Issue #511, opened 2025-10-09
Problems with connection after few days


treghart:
Hello,

I have a strange problem. I am using this integration with T4A-PV and newest HA. It connected fine, I am able to run automations on human detect etc. But after few days cameras loose their connection to HA, and I can't even login through www directly to the camera as it says that number of connections reached its limits. I need to restart camera for it to work again. It shows fine on NVR though. I am wondering if somehow integration might not be closing connections properly so they stay alive and reach limits? Or it could be something else?

zios0007:
Resolved, as @rroller says: the change that caused it was reverted in 0.9.75, and @treghart confirmed.

Worth recording what the symptom was, because it is unusual and someone will hit something like it again: the camera ran out of connection slots, so even its own web interface refused to log in until it was restarted. That is a device with a small connection limit meeting something that opened more than it closed, which is why a revert fixed it rather than a setting.

Two things since then reduce that pressure generally: the connector is shared per host rather than one per entry, so several channels on one device hold one rather than several, and a failed setup gives its connection back instead of leaking one per retry.

@rroller closeable.

PR #597, opened 2026-08-30
Reuse the digest challenge instead of re-authenticating on every request

zios0007:
Fixes the "massive login/logout" behaviour reported in #577, and very likely the connection exhaustion in #511.

## The bug

`DigestAuth.__init__` takes a `previous=` argument so an accepted challenge can be carried into the next request. None of the three call sites in `client.py` ever passes it:

```python
auth = DigestAuth(self._username, self._password, self._session)
```

So `self.challenge` is always `None`, every API call goes out unauthenticated, takes a `401`, and is retried. **Two HTTP requests per logical call**, forever. The `previous` parameter has been dead code the whole time — nothing read it, nothing wrote back to it.

That is cheap on a device that supports keep-alive. It is not cheap here: my NVR answers `Connection: close` on every response, including the 401. Keep-alive is impossible, so each of those requests is its own TCP connection, TLS handshake, and **login**.

## Measurements

DHI-NVR5464-16P-EI, 12 channels configured, HTTPS on 443. Sampling the NVR's own `userManager.cgi?action=getActiveUserInfoAll` every 2 minutes:

```
mean 2.13 logins/sec  ->  7,654 logins/hour
concurrent sessions held by HA: 10-22
```

For a controlled comparison, I rebooted the NVR and caught an 8-minute window before this integration reconnected. Same device, same other clients:

| | logins/sec |
|---|---|
| everything except this integration | 0.18 – 0.24 |
| after it reconnected | **2.13** |

The device's CGI service degrades as this accumulates — response times went from ~130 ms to 3 s within 100 minutes of boot — and eventually stops answering HTTP entirely while still completing TCP connects. Only a reboot clears it. That matches #511's *"number of connections reached its limits"*, and it takes P2P/DMSS remote access down with it, since remote access is the client that must open a *new* session on demand.

@dawuhe's numbers in #577 line up: 114,300 logins in 24 h from 5 cameras is 1.32/s, against my 2.13/s from 12 channels. Same per-channel rate, different channel counts.

## What this changes

**1. The client keeps one challenge.** `DahuaClient` holds a single `_digest_state` dict and passes it at all three call sites. `DigestAuth` holds it by reference and exposes `challenge` / `last_nonce` / `nonce_count` as properties backed by it, so the rest of the class is untouched. Callers that pass nothing get their own private state and the previous behaviour, so this is not a breaking change for any other caller.

The nonce counter now advances across requests as RFC 2617 intends, rather than every request minting `nc=00000001` on its own private counter. The read-modify-write is synchronous with no `await` in it, so concurrent requests get distinct counts.

**2. The retry is bounded.** `_handle_401` recursed through `request()` with no limit. A device that rejects credentials returns 401 *with* a `www-authenticate` header, so the old code re-challenged forever. Against current `main`:

```
upstream HEAD  ->  RecursionError after 495 HTTP requests
this branch    ->  status 401 after 2
```

495 logins to the device for one API call. If any of the entries in #577 came from a config entry with stale credentials, that alone would account for a login count far above what 2x amplification explains.

The budget distinguishes the two causes of a 401 on a cached challenge: a rotated nonce or `stale=true` is retried, while the same nonce returned twice means the password is wrong and we stop. One same-nonce retry is allowed first, so a device that refuses an out-of-order `nc` per RFC 7616 §5.5 recovers instead of failing the call.

**3. A bad cached challenge no longer sticks.** `_build_digest_header` did `self.challenge["realm"]`. With a cached challenge that is now reused, one malformed 401 during a device reboot would raise `KeyError` on every subsequent request for the life of the config entry. It now returns `""`, and `request()` drops the challenge and re-probes.

## Tests

`tests/dahua/test_digest.py`, 9 tests. The fake device verifies the digest it is sent — realm, nonce, `nc`, and the MD5 response — rather than just checking a header exists, and it suspends before replying so gathered requests genuinely overlap instead of each running to completion in one scheduler step.

Covered: challenge reuse across calls; the client threading one challenge through all three call sites; unchanged behaviour when no state is passed; `nc` advancing; distinct `nc` under real concurrency; recovery from a refused out-of-order `nc`; bounded retry on bad credentials; nonce rotation refreshed *and* the retry succeeding; an unusable cached challenge being discarded.

I checked the client-level test actually pins the fix — dropping the fourth argument from the three call sites fails it.

## Effect

For a poll cycle that currently costs N connections, this costs N/2 in the steady state. It does not change how many API calls the coordinator makes per cycle, so it halves the connection and handshake load rather than eliminating it — on this device the remaining login rate is still driven by `SCAN_INTERVAL_SECONDS` (hardcoded to 30) multiplied by the number of configured channels, each of which gets its own coordinator, `ClientSession`, and permanently-open event stream. Those are worth looking at separately; I wanted to keep this PR to the auth path.

Happy to split the retry-bounding into its own PR if you'd rather take that first — it stands alone and is the smaller change.

### DIFF
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -35,6 +35,9 @@ class DahuaClient:
     ) -> None:
         self._username = username
         self._password = password
+        # One digest challenge shared by every request this client makes, so a
+        # call doesn't have to take a 401 before it can authenticate.
+        self._digest_state = {}
         # Strip trailing slashes from address to prevent malformed URLs like http://host/:80
         self._address = address.rstrip('/')
         self._session = session
@@ -773,7 +776,7 @@ class DahuaClient:
             response = None
 
             try:
-                auth = DigestAuth(self._username, self._password, self._session)
+                auth = DigestAuth(self._username, self._password, self._session, self._digest_state)
                 response = await auth.request("GET", url)
                 response.raise_for_status()
 
@@ -814,7 +817,7 @@ class DahuaClient:
         async with async_timeout.timeout(TIMEOUT_SECONDS):
             response = None
             try:
-                auth = DigestAuth(self._username, self._password, self._session)
+                auth = DigestAuth(self._username, self._password, self._session, self._digest_state)
                 response = await auth.request("GET", self._base + url)
                 response.raise_for_status()
 
@@ -830,7 +833,7 @@ class DahuaClient:
             async with async_timeout.timeout(TIMEOUT_SECONDS):
                 response = None
                 try:
-                    auth = DigestAuth(self._username, self._password, self._session)
+                    auth = DigestAuth(self._username, self._password, self._session, self._digest_state)
                     response = await auth.request("GET", url)
                     response.raise_for_status()
                     data = await response.text()
--- a/custom_components/dahua/digest.py
+++ b/custom_components/dahua/digest.py
@@ -12,6 +12,9 @@ from yarl import URL
 # Copied and then modified from https://github.com/aio-libs/aiohttp/pull/2213
 # I really wish this was baked into aiohttp :-(
 
+# How many times one request may answer a 401 before giving up.
+MAX_AUTH_ATTEMPTS = 3
+
 
 class DigestAuth:
     """HTTP digest authentication helper.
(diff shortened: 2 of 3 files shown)


rroller:
This one has merge conflicts

zios0007:
Rebased onto current `main` — the conflict was in the event-stream request, where #598 bounded the long poll on socket read while this branch threads shared digest state through `DigestAuth`. The two are orthogonal, so the resolution keeps both the read timeout and the reused challenge.

The `Run tests` check is red for reasons unrelated to this PR. `tests/dahua/test_issue589.py` has never run in CI — it fails at collection on `ModuleNotFoundError: turbojpeg` — and once that is fixed, three further pre-existing failures appear. #598 was itself merged with `Run tests` already failing.

I've opened #599 with those fixes, deliberately separate so this PR stays a focused change. Verified in a `python:3.13` container matching CI:

```
this PR alone      : 2 failed, 18 passed, 7 errors   (all failures pre-exist on main)
this PR + #599     : 27 passed
this PR's own tests: 9 passed  (tests/dahua/test_digest.py)
```

So merging #599 first will turn this check green with no further changes here.

rroller:
The tests are failing FYI

rroller (approved):

PR #600, opened 2026-08-31
Send the channel number, not the channel index, to ptz.cgi

### LINKED ISSUE #522: PTZ Preset not working from automation or nodered
When I execute the following flow nothing happens...

The flow triggers, the message appears to be sent - the camera does not repoint.

I'm not sure where to from here on this one.

Image

Image

Image

### LINKED ISSUE #592: PTZ presets off by one on NVR: select.py uses get_channel() instead of get_channel_number()
PTZ preset uses channel index instead of channel number (off-by-one on NVR)

select.py line 86 calls self._coordinator.get_channel(), which the docstring defines as "the channel index of this camera. 0 based". The PTZ CGI expects the 1-based channel number. camera.py line 245 already uses get_channel_number() for the RTSP URL.

Result on an NVR: video is correct but GotoPreset moves the camera on the previous channel. With gaps in the channel list, the call instead returns 400 Bad Request.

Measured on an 8-channel NVR with cameras on D1/D3/D4: the entity for channel index 2 produced ptz.cgi?channel=2 → 400, while the entity for index 3 physically moved the camera on D3.

Changing line 86 to get_channel_number() fixes it. Version 0.9.83.

zios0007:
Fixes #592. Very likely also fixes #522.

`get_channel()` is documented as *"the channel index of this camera. 0 based"*, but `ptz.cgi` expects the channel **number**. `camera.py` already uses `get_channel_number()` for the RTSP URL — the two PTZ preset paths did not.

On an NVR that means a preset moves the camera on the wrong channel, or returns `400 Bad Request` when the channel list has gaps.

### Two call sites, not one

#592 reports the select entity (`select.py`). The `goto_preset_position` **service** in `camera.py` has the identical defect and isn't mentioned there — it takes `self._logical_channel`, which is `get_channel()`. That is the likely cause of #522, where a preset called from an automation or Node-RED does nothing. Both are fixed here.

### Why `get_channel_number()` and not `index + 1`

The coordinator defaults the number to `channel + 1`, but auto-detects older firmware that numbers channels from zero and resets the number back to the index:

```python
# __init__.py
self._channel_number = channel + 1
...
    if not self.is_doorbell():
        self._channel_number = self._channel
```

Computing `index + 1` at the call site would silently undo that correction. Reading `get_channel_number()` keeps it. `test_channel_number_equal_to_index_is_respected` guards this.

### Verification

In a `python:3.13` container matching the CI job, running the new tests against unmodified code and then against this change:

```
without the fix: 2 failed, 2 passed   (channel 2 sent where 3 was required)
with the fix:    4 passed
```

### Note on CI

`Run tests` will be red on this branch for reasons unrelated to it — `tests/dahua/test_issue589.py` fails at collection on `ModuleNotFoundError: turbojpeg`, and the resulting event-loop damage cascades into other tests. #599 fixes that. With #599 merged in locally, the full suite including these new tests is **22 passed**.

`async_get_ptz_preset_ids` is deliberately left alone — it talks to RPC2, not `ptz.cgi`, and its parameter is correctly named `channel_index`.

### DIFF
--- a/custom_components/dahua/camera.py
+++ b/custom_components/dahua/camera.py
@@ -350,7 +350,7 @@ class DahuaCamera(DahuaBaseEntity, Camera):
 
     async def async_goto_preset_position(self, position: int):
         """Go to a preset, using RPC2 only for the SDT4E425."""
-        channel = self._logical_channel
+        channel = self._channel_number
         if is_sdt4e425(self._coordinator.get_model()):
             await self._coordinator.client.async_goto_preset_rpc2(1, position)
         else:
--- a/custom_components/dahua/select.py
+++ b/custom_components/dahua/select.py
@@ -111,7 +111,7 @@ class DahuaCameraPresetPositionSelect(DahuaBaseEntity, SelectEntity):
                 self._rpc2_channel, int(option)
             )
         else:
-            channel = self._coordinator.get_channel()
+            channel = self._coordinator.get_channel_number()
             await self._coordinator.client.async_goto_preset_position(channel, int(option))
         await self._coordinator.async_refresh()
 
--- a/tests/dahua/test_ptz_channel.py
+++ b/tests/dahua/test_ptz_channel.py
@@ -1,0 +1,100 @@
+"""ptz.cgi takes the channel number, not the 0-based channel index."""
+
+from types import SimpleNamespace
+
+from custom_components.dahua.camera import DahuaCamera
+from custom_components.dahua.select import DahuaCameraPresetPositionSelect
+
+
+class _CapturingClient:
+    """Records the channel handed to the PTZ CGI call."""
+
+    def __init__(self):
+        self.channel = None
+        self.position = None
+
+    async def async_goto_preset_position(self, channel, position):
+        self.channel = channel
+        self.position = position
+        return {}
+
+
+class _FakeCoordinator:
+    def __init__(self, channel_index, channel_number, model="OTHER"):
+        self.client = _CapturingClient()
+        self._channel_index = channel_index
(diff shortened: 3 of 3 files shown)


zios0007:
To make the merge order unambiguous: this PR's tests are **hard-blocked** by #599, not merely coexisting with the breakage on `main`.

`tests/dahua/test_ptz_channel.py` imports `custom_components.dahua.camera`, which imports `homeassistant.components.camera`, which imports `turbojpeg`. `PyTurboJPEG` is not in `requirements_test.txt`, so the module cannot be collected at all:

```
ERROR collecting tests/dahua/test_ptz_channel.py
ModuleNotFoundError: No module named 'turbojpeg'
no tests collected, 1 error
```

That is the whole of this PR's contribution to the red check — the other two entries (`test_event_stream.py`, `test_issue589.py`) are pre-existing on `main`.

Confirmed locally by removing `PyTurboJPEG` from a `python:3.13` container matching CI, which reproduces the error exactly, and reinstalling it, which gives `4 passed`.

This generalises: **any** test that touches the camera platform is uncollectable today. That is likely why `test_issue589.py` is the only camera test in the repo and why it has never actually run.

Suggested order: merge #599 first, then this and #597 need only a `main` merge. Verified locally with #599 merged in:

```
#599 alone     : 18 passed
#600 + #599    : 22 passed
#597 + #599    : 27 passed
```

Happy to rebase this onto #599 instead if you would rather see a green check here before reviewing — it just means this PR temporarily carries #599's commit as well.

rroller (approved):

PR #602, opened 2026-08-31
Make the device poll interval configurable

zios0007:
The coordinator polls every 30 seconds, hardcoded:

```python
SCAN_INTERVAL_SECONDS = timedelta(seconds=30)
...
update_interval=SCAN_INTERVAL_SECONDS,
```

Each cycle fans out to one unconditional CGI call plus up to eight capability-dependent ones (`async_get_config_lighting`, `async_get_disarming_linkage`, `async_get_event_notifications`, `async_get_coaxial_control_io_status`, `async_get_smart_motion_detection`, …), plus `async_get_video_in_mode` and `async_get_ptz_position` — **per entry**.

That is a lot of traffic for what it actually reads. Events do not come from polling; they arrive on the event stream. The poll only refreshes configuration state — motion detection settings, disarming linkage, event notification settings, lighting mode — and those change rarely.

It compounds on an NVR, where each channel is a separate config entry against the same host. On an eleven-entry setup, 30-second polling is on the order of a hundred thousand requests a day, and there is currently no way to slow it down short of editing the source. It is likely a contributor to #577.

### What this does

Adds the interval to the options screen:

```python
def get_configured_scan_interval(entry: ConfigEntry) -> timedelta:
    seconds = entry.options.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL)
    ...
    return timedelta(seconds=max(seconds, MIN_SCAN_INTERVAL))
```

- Defaults to the previous 30 seconds, so nothing changes for existing entries.
- The form refuses anything below `MIN_SCAN_INTERVAL` (10s), and the read side clamps as well, so a hand-edited entry cannot hammer the device.
- A non-numeric value falls back to the default rather than raising during setup.
- The entry already reloads on an options update via the existing update listener.
- `SCAN_INTERVAL_SECONDS` is removed as it had no remaining references.

### Verification

```
options screen reverted, tests kept : 3 failed, 6 passed  (KeyError: 'scan_interval')
with this change                    : 9 passed
full suite with #599 merged         : 27 passed
```

These tests import only `custom_components.dahua`, its `const`, and `config_flow` — none of which pull in `homeassistant.components.camera` — so they collect and run on CI without needing #599 first.

Note the `Run tests` check will still be red here, from the two failures that pre-exist on `main` and that #599 fixes. This PR's own tests pass within that run.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -38,11 +38,13 @@ from .const import (
     STARTUP_MESSAGE,
     CONF_CHANNEL,
     CONF_AUTO_DETECT_CHANNEL,
+    CONF_SCAN_INTERVAL,
+    DEFAULT_SCAN_INTERVAL,
+    MIN_SCAN_INTERVAL,
 )
 from .dahua_utils import parse_event
 from .vto import DahuaVTOClient
 
-SCAN_INTERVAL_SECONDS = timedelta(seconds=30)
 
 # A stream that keeps heartbeating but has stopped reporting events looks
 # healthy to a read timeout, so recycle it periodically as well.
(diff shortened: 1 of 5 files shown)


rroller (approved):

rroller:
This one has merge conflicts, but otherwise, feel free to merge it

zios0007:
Thanks — the conflicts were with #601, which had merged a couple of minutes earlier. Both PRs appended to the same three places: the options schema in `config_flow.py`, the helper above `async_setup_entry` in `__init__.py`, and the label list in `en.json`. They are independent settings, so the resolution keeps both.

Already pushed and green. Verified in a `python:3.13` container matching CI: 46 passed, which is everything from #597, #599, #600 and #601 plus this PR's 9.

Merging as you suggested.

rroller (approved):

rroller:
sorry this one has merge conflicts

rroller (approved):

PR #607, opened 2026-09-03
Stop fetching what the integration does not use

zios0007:
Four wasted requests, found while looking at why a busy NVR struggles. All are independent of each other.

**Builds on #606** — it shares that branch, so the diff currently shows its commit too. Once #606 merges this collapses to its own change.

### 1. A full JPEG downloaded per entry, to answer a yes/no question

```python
await self.client.async_get_snapshot(0)   # → get_bytes → await response.read()
```

This runs at every entry setup when `auto_detect_channel` is on, purely to see whether the endpoint answers. The image is discarded. On an NVR with one entry per channel that is a simultaneous full-resolution download per entry, at the moment the device is already handling one CGI call per entry.

`async_probe_snapshot` issues the same request against the same endpoint — so the heuristic is unchanged — and releases the response with `close()` instead of `read()`, which drops the body rather than transferring it.

### 2. A new `ClientSession` for every PTZ call

`_new_rpc2_session()` was called per request, building a session, a connector and a connection pool each time and discarding them. Sessions are meant to be long lived. It is now built once per client and closed alongside the existing session teardown in `_close_session`.

### 3. `async_get_lighting_v2` requested twice per poll

Once in the fan-out when `_supports_lighting_v2` (set at runtime when the API answers), and again afterwards for `supports_security_light() or is_flood_light()` (decided by model name). Those overlap — an `-AS-PV` camera hits both — so the same call went out twice on every single poll. The trailing call now runs only when it was not already in the fan-out.

### 4. One avoidable serial round trip per poll

`async_get_ptz_position` was awaited on its own ahead of the fan-out, though nothing in the batch depends on it. It joins the batch, keeping the same swallow-and-log behaviour via a small wrapper.

`async_get_video_in_mode` above it is **not** moved: it sets `self._profile_mode`, which `async_get_config_lighting` then consumes, so it is a real dependency.

### Verification

```
suite: 73 passed
probe test with the body download reintroduced: 1 failed
```

The new tests cover the probe not reading the body, `get_bytes` still reading it, and the RPC2 session being built once, released on close, replaced if closed, and safe to close twice.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -277,6 +277,10 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
 
     async def _close_session(self) -> None:
         _LOGGER.debug("Closing Session")
+        try:
+            await self.client.close()
+        except Exception:
+            _LOGGER.debug("Failed to close the client's RPC2 session", exc_info=True)
         if self._session is not None:
             try:
                 await self._session.close()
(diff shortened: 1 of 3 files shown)


rroller (approved):

rroller:
this one has a conflict, feel free to resolve and merge

PR #609, opened 2026-09-03
Turn lights on at full when no brightness is given

zios0007:
`light.py` had no tests at all — 211 statements at 0% coverage. Writing the first ones turned up a bug that has been there since the initial commit.

**Builds on #606, #607 and #608** and shares their branch, so the diff currently shows those commits too. It collapses to its own change once they merge.

### The bug

The two brightness conversions both default to `100`, but they work on different scales. Full is **100** on the Dahua scale and **255** on the HASS scale:

```python
def dahua_brightness_to_hass_brightness(bri_str):   # Dahua -> HASS
    bri = 100                                       # -> 255, full, correct

def hass_brightness_to_dahua_brightness(hass_brightness):   # HASS -> Dahua
    if hass_brightness is None:
        hass_brightness = 100                       # -> 39, meant to be full
```

Home Assistant omits `ATTR_BRIGHTNESS` when a light is simply toggled on, so **every plain turn-on of an infrared light or an illuminator asked the camera for 39% brightness rather than full.** All four call sites in `light.py` pass `kwargs.get(ATTR_BRIGHTNESS)`, so they all reach it.

Defaulting to `255` gives `int((255/255)*100) == 100` — full on the scale this function returns, and consistent with its sibling.

### The tests

Aimed at the parts most likely to regress rather than at a coverage number:

- **The illuminator passes `profile_mode` through** to `async_set_lighting_v2` on both on and off, and reflects whatever the coordinator currently reports. Day/night handling rides on that argument, and it is what #582 got wrong badly enough to be reverted.
- **The two APIs never cross** — the infrared light must use v1 and the illuminator v2. Each test asserts the other was not called, so a copy-paste between the two classes fails loudly.
- **Unique IDs stay distinct** (`_infrared` vs `_illuminator`); a collision would silently merge two lights into one entity.
- **Conversion boundaries** — 0, 255, and the `None` case above, plus a round-trip check.

They construct the entities with `object.__new__` and a fake coordinator, so no `hass` fixture is needed. The same approach would work for `switch.py` (128 statements) and `binary_sensor.py` (55), which are also at zero.

### Verification

```
suite: 99 passed
light.py       0%  -> 59%
dahua_utils.py 74% -> 97%
TOTAL          36% -> 41%
```

Reintroducing the old default fails three tests.

### Note

This changes observable behaviour: lights that came on dim will now come on at full. That is the intent of the code as documented, but it is a visible change and worth calling out.

### DIFF
--- a/custom_components/dahua/dahua_utils.py
+++ b/custom_components/dahua/dahua_utils.py
@@ -23,7 +23,9 @@ def hass_brightness_to_dahua_brightness(hass_brightness: int) -> int:
     Converts a HASS brightness (which is 0 to 255 inclusive) to a Dahua brightness (which is 0 to 100 inclusive)
     """
     if hass_brightness is None:
-        hass_brightness = 100
+        # No brightness asked for means full. Full on the HASS scale is 255;
+        # 100 here is a Dahua-scale value and came out as 39% instead.
+        hass_brightness = 255
     return int((hass_brightness / 255) * 100)
 
 
--- a/tests/dahua/test_light.py
+++ b/tests/dahua/test_light.py
@@ -1,0 +1,194 @@
+"""light.py had no tests at all, and day/night handling has regressed before."""
+
+import pytest
+
+from custom_components.dahua import dahua_utils
+from custom_components.dahua.light import DahuaIlluminator, DahuaInfraredLight
+
+ATTR_BRIGHTNESS = "brightness"
+
+
+class _Client:
+    def __init__(self):
+        self.v1 = []
+        self.v2 = []
+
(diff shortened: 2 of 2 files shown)


rroller (approved):

rroller:
also a conflic,t feel free to merge once resolved

zios0007:
Rebased directly onto `main`, so this no longer depends on #607 or #608 and can be merged in any order. It only touches `dahua_utils.py` and adds `tests/dahua/test_light.py`, so there was no real dependency on either — it was just where the branch happened to sit.

Suite is 87 passed on this branch alone.

PR #610, opened 2026-09-03
Add tests for switch.py and binary_sensor.py

zios0007:
The last two platforms with no coverage at all. **No bugs turned up this time** — unlike #609, where the first `light.py` test failed immediately. These pin behaviour that is easy to break silently.

**Builds on #606, #607, #608 and #609** and shares their branch, so the diff currently shows those commits too. It collapses to its own change once they merge.

### switch.py

Each switch drives a different API, and the arguments are what separate them:

- **The siren and the security light both go through `async_set_coaxial_control_state`** and differ only by a type constant. A swapped constant would sound the siren when the light was asked for, with nothing else to catch it. Two tests assert `SIREN_TYPE` specifically, on and off.
- **Smart motion picks between the Amcrest IVS rule and the Dahua API by vendor** — `async_set_ivs_rule(0, 0, x)` versus `async_enabled_smart_motion_detection(x)`. Both branches are covered.
- **The three channel-scoped switches** are asserted to send the right channel and the right boolean, and to refresh afterwards — without which the UI shows a stale state.
- **Every switch keeps its own `unique_id`.** A collision merges two switches into one entity.

### binary_sensor.py

Names, device classes and unique IDs are all *derived* rather than declared:

- **The name derivation** turns `SmartMotionHuman` into "Smart Motion Human" while leaving consecutive capitals like `IVS` alone — that lookbehind is easy to break with a well-meaning regex tidy.
- **Overrides win over derivation** for `VideoMotion`, `CrossLineDetection` and `DoorbellPressed`.
- **`VideoMotion` deliberately keeps the bare serial number as its unique ID** for backwards compatibility. The comment says so; now a test does too, because tidying it away would orphan every existing motion sensor on upgrade.
- **IDs stay distinct** across the nine event codes a typical camera reports.
- **State follows the event timestamp**, and a sensor does not react to a different event's timestamp.

### Verification

```
suite: 136 passed

switch.py         0% -> 85%
binary_sensor.py  0% -> 80%
total            41% -> 48%
```

Pointing the siren at `SECURITY_LIGHT_TYPE` fails two tests.

These use the same `object.__new__` plus fake-coordinator approach as #609, so no `hass` fixture is needed and they run fast.

### DIFF
--- a/custom_components/dahua/dahua_utils.py
+++ b/custom_components/dahua/dahua_utils.py
@@ -23,7 +23,9 @@ def hass_brightness_to_dahua_brightness(hass_brightness: int) -> int:
     Converts a HASS brightness (which is 0 to 255 inclusive) to a Dahua brightness (which is 0 to 100 inclusive)
     """
     if hass_brightness is None:
-        hass_brightness = 100
+        # No brightness asked for means full. Full on the HASS scale is 255;
+        # 100 here is a Dahua-scale value and came out as 39% instead.
+        hass_brightness = 255
     return int((hass_brightness / 255) * 100)
 
 
--- a/tests/dahua/test_binary_sensor.py
+++ b/tests/dahua/test_binary_sensor.py
@@ -1,0 +1,145 @@
+"""binary_sensor.py had no tests. Its naming and IDs are derived, not declared."""
+
+import pytest
+
+from custom_components.dahua import binary_sensor as bs
+from custom_components.dahua.binary_sensor import DahuaEventSensor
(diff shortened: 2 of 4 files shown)


rroller (approved):

zios0007:
Rebased so this now sits directly on #609, which in turn sits on `main` — neither depends on #607 or #608 any more. This one only adds test files, so it can be merged whenever #609 is.

Suite is 124 passed on this branch.

rroller (approved):

Issue #555, opened 2026-03-11
Unable to trigger siren on DH-P5AE-PV – `coaxialControlIO.cgi` returns Bad Request


toloomierz:
### Describe the issue

I have a Dahua DH-P5AE-PV camera (wireless Pan/Tilt/Zoom from the Active Deterrence series) running firmware 2.820.0000016.0.R (build date 2025-06-17). I am using the dahua integration via Home Assistant but also tried triggering the built-in siren directly via the HTTP API.

According to the Dahua HTTP API documentation, the siren/light should be controlled via `coaxialControlIO.cgi?action=control` with `Type=2` for the siren, `IO=1` to turn it on and an optional `TriggerMode=2` for manual trigger. Example:

```bash
curl --globoff --digest -u USER:PASS \
  "http://<camera_ip>/cgi-bin/coaxialControlIO.cgi?action=control&channel=1&info[0].Type=2&info[0].IO=1&info[0].TriggerMode=2"
```

However, this camera always returns:

```
Error
Bad Request!
```

I have tried changing the `channel` parameter (0, 1, 2) and the `TriggerMode` (0, 1, 2), but the result is always a 400 `Bad Request` with "Error Bad Request!" in the body. The white light parameter (`Type=1`) returns the same error. Digest authentication is working correctly (other endpoints, such as `System/information`, return valid JSON).

### Expected behaviour

The command above should activate the built-in siren for ~10 s and return `OK` in the response body as documented for other models.

### Steps to reproduce

1. Add a DH-P5AE-PV camera with digest auth.
2. Run the `curl` GET request shown above.
3. Observe the 400/`Bad Request` response.

### Additional context

- The siren and light can be activated through the camera's DMSS mobile app, so the hardware works.
- Another user on ipcamtalk forums reported the same issue on different models – the command works on some cameras but returns `Bad Request` on others.

Is there an alternative endpoint for the DH-P5AE-PV or is this camera unsupported by the `coaxialControlIO.cgi` API?

Any guidance would be appreciated.

zios0007:
@toloomierz your diagnosis is sound and the answer is probably one you will not like: on that camera, the siren is very likely not reachable over CGI at all, and there is nothing the integration can do about it.

Three things support that.

**You have already ruled out the integration.** Your `curl` bypasses it entirely and still returns `400`, with digest auth demonstrably working on other endpoints. Whatever this is, it is between the camera and the CGI API.

[...]

Issue #546, opened 2026-02-22
[Bug] TiOC 3 / TiOC Pro - Security Light (light entity) stays ON after turn_off command


lsantoro:
I am experiencing a synchronization issue with the latest Dahua TiOC 3 and TiOC Pro models (unlike TiOC 2 models which work perfectly).
When I trigger the security_light(recognized as a light.xxx entity) from Home Assistant:
1. Turn On: The light turns on correctly.
2. Turn Off: The entity in Home Assistant updates to "off", but the physical light on the camera remains ON.
Key Observations:
* App Discrepancy: When the light is stuck ON (triggered by HA), the Dahua DMSS app shows the light as "Disabled". To turn it off, I must manually enable it in DMSS and then disable it again.
* Siren Works: The siren (recognized as a switch.xxx entity) works perfectly for both ON and OFF commands.
* Model Specific: This issue only occurs on TiOC 3/Pro series. TiOC 2 models on the same network/integration do not exhibit this behavior.
Environment:
* Integration: Dahua  0.9.76
* Camera Models: DH-IPC-HFW3849T1-AS-PV Firmware: 3.140.0000000.43.R,build:2025-12-12 DH-IPC-HFW3849T1-AS-PV-PRO Firmware: 3.140.0000000.21.R,build:2025-12-06
* Home Assistant Version: 2026.2.3

zios0007:
@lsantoro your report has aged into something useful: it is now one of several on the **3.140** firmware generation, and that generation has turned out to handle lighting differently from everything before it.

Three independent reports on #647 share it — `V3.140.0000000.33.R` (DH-IPC-HFW3449E-S-IL), `3.140.0000000.81.R` (IPC-HDW3849HP-AS-PV) — and your TiOC 3 / Pro cameras are the same family. Your own observation that **TiOC 2 works perfectly on the same network and integration** fits that exactly.

**The detail in your report that now looks important** is the one that would have been easy to dismiss:

> When the light is stuck ON (triggered by HA), the Dahua DMSS app shows the light as "Disabled".

The camera's own app says off while the light is physically on. That is the same divergence @roby-esc measured directly on #647: they watched Smart Dual Light switch physically between infrared and white light **with no change in `VideoInMode`, `Lighting`, `Lighting_V2` or `LightingScheme`**. Their conclusion is the sentence I would apply to your case too — *these tables describe configuration, not the current physical light state.*

[...]

Issue #446, opened 2025-01-23
Security light and siren entity issues with Dahua TiOC camera


BM183:
Hi,

I have several DH-IPC-HDW3649H-AS-PV-ANZ cameras, which work quite well with the integration. I'm on the latest FW 3.140.0000000.24.R,build:2024-07-04.

One issue, is the use of the 'security light' and 'siren' controls don't register/ feedback the state of the entity when used. So switching the light on does not keep that state in HA, the toggle returns to off by itself even though the light is actually activated and in the dahua web ui live view shows as activated and on. This causes some issues with automations, and also means it's not showing in the device logbook.

Strangely, other control entities like illuminator work just fine and detect state.

Any ideas on how to get this fully working? Can post a log if it helps, but need guidance as to what to filter specifically. Thanks!

zios0007:
@BM183 your report has turned out to be the **earliest firmware 3.140 sighting** in this repo, and it went unanswered for twenty months. It is worth connecting, because there are now several of them and they look like one problem.

You wrote `3.140.0000000.24.R` in January 2025. Since then:

| Issue | Camera | Firmware |
|---|---|---|
| this one | DH-IPC-HDW3649H-AS-PV | `3.140.0000000.24.R` |
| #546 / #494 | TiOC 3 / TiOC Pro | 3.140 generation |
| #647 | DH-IPC-HFW3449E-S-IL | `V3.140.0000000.33.R` |
| #647 | IPC-HDW3849HP-AS-PV | `3.140.0000000.81.R` |

Your specific symptom — *the toggle returns to off by itself even though the light is actually activated* — is the one @roby-esc explained on #647 by watching the hardware rather than the API. They measured Smart Dual Light switching physically between infrared and white light **with no change in `VideoInMode`, `Lighting`, `Lighting_V2` or `LightingScheme`**.

That is exactly your case read from the other direction: Home Assistant refreshes its state from those tables, the tables say the light is off, so the toggle springs back — while the light is on and the Dahua UI agrees with you rather than with the integration. The integration is not losing the command; it is reading a source that does not describe the physical light on this firmware.

**I am not claiming that is solved.** Nobody has found a value on these cameras that reports the live lighting state — #647 went through `SceneModeManager`, `devVideoInput.getStatus` and `getVideoInStatus`, and ruled all of them out. So state feedback for 3.140 lighting remains genuinely open, and your report is the oldest evidence for it.

[...]

PR #611, opened 2026-09-04
Add NVR TiOC active deterrence control and status synchronization

Saufziege:
## Description

This pull request adds reliable coaxial control and status synchronization for TiOC cameras connected through Dahua NVR channels.

Dahua NVRs mask the capabilities of connected cameras, so Home Assistant could fail to discover the active deterrence entities and could send light or siren commands through the wrong API.

These changes:

- Force the creation of the `Warning Light` and `Alarm` entities for NVR channels.
- Route NVR warning-light commands through `coaxialControlIO.cgi` with `Type=1`.
- Route NVR siren commands through `coaxialControlIO.cgi` with `Type=2`.
- Map `IO=1` to ON and `IO=2` to OFF.
- Include `TriggerMode=2` for NVR control requests.
- Use the resolved NVR channel number for control and status requests.
- Preserve the original direct-camera behavior and keep the normal `Illuminator` separate from the `Warning Light`.

## Tested Hardware

- **NVR:** DHI-NVR4116HS-8P-EI
- **Camera:** DH-IPC-HFW3449T1-AS-PV-S3
- **Camera type:** TiOC 3.0 / WizSense S3

## Verification

Warning-light and siren ON/OFF control was tested successfully. Status and event history in Home Assistant remained synchronized with the physical device and the DMSS app.
Screenshot 2026-09-04 122151

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -45,6 +45,7 @@ from .const import (
     CONF_AUTO_DETECT_CHANNEL,
     CONF_USE_HTTPS,
     CONF_SCAN_INTERVAL,
+    CONF_NVR_ACTIVE_DETERRENCE,
     DEFAULT_SCAN_INTERVAL,
     MIN_SCAN_INTERVAL,
 )
@@ -623,6 +624,7 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         self.connected = None
         self.events: list = events
         self._supports_coaxial_control = False
+        self._nvr_active_deterrence = entry.options.get(CONF_NVR_ACTIVE_DETERRENCE, False)
         self._supports_disarming_linkage = False
         self._supports_event_notifications = False
         self._supports_smart_motion_detection = False
@@ -1223,6 +1225,10 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         m = self.model.upper()
         return "-AS-PV" in m or "L46N" in m or m.startswith("W452ASD")
 
+    def supports_nvr_active_deterrence(self) -> bool:
+        """Return whether NVR active-deterrence entities were explicitly enabled."""
+        return self._nvr_active_deterrence
+
     def supports_security_light(self) -> bool:
         """
         Returns true if this camera has the red/blue flashing security light feature.  For example, the
@@ -1524,7 +1530,13 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
 
 async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     """Handle removal of an entry."""
-    coordinator = hass.data[DOMAIN][entry.entry_id]
+    coordinator = hass.data.get(DOMAIN, {}).get(entry.entry_id)
+    if coordinator is None:
+        # Setup may have failed before the coordinator was registered, or a
+        # previous unload may already have removed it. Treat that as unloaded
+        # so an options-triggered reload can continue cleanly.
+        return True
+
     await coordinator.async_stop()
     unloaded = all(
         await asyncio.gather(
@@ -1553,5 +1565,4 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
 async def async_reload_entry(hass: HomeAssistant, entry: ConfigEntry) -> None:
     """Reload config entry."""
-    await async_unload_entry(hass, entry)
-    await async_setup_entry(hass, entry)
+    await hass.config_entries.async_reload(entry.entry_id)
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -1463,6 +1463,10 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         """returns the channel index of this camera. 0 based. Channel index 0 is channel number 1"""
         return self._channel
 
+    def is_nvr_channel(self) -> bool:
+        """Return whether this entry represents a camera channel on an NVR."""
+        return self._channel > 0 or "NVR" in self.model.upper()
+
     def get_channel_number(self) -> int:
         """returns the channel number of this camera"""
         return self._channel_number
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -836,7 +836,8 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                 _LOGGER.debug("Using channel number %s (auto_detect=%s)", self._channel_number, auto_detect)
 
                 try:
-                    await self.client.async_get_coaxial_control_io_status()
+                    coaxial_channel = self._channel_number if self.is_nvr_channel() else 1
+                    await self.client.async_get_coaxial_control_io_status(coaxial_channel)
                     self._supports_coaxial_control = True
                 except ClientResponseError:
                     self._supports_coaxial_control = False
@@ -979,7 +980,12 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
                 coros.append(asyncio.ensure_future(self.client.async_get_event_notifications()))
             # The siren switch and the security light both read this one.
             if self._supports_coaxial_control and self._wanted_by(LIGHT, SWITCH):
-                coros.append(asyncio.ensure_future(self.client.async_get_coaxial_control_io_status()))
+                coaxial_channel = self._channel_number if self.is_nvr_channel() else 1
+                coros.append(
+                    asyncio.ensure_future(
+                        self.client.async_get_coaxial_control_io_status(coaxial_channel)
+                    )
+                )
             if self._supports_smart_motion_detection and self._wanted_by(SWITCH):
                 coros.append(asyncio.ensure_future(self.client.async_get_smart_motion_detection()))
             if self.supports_smart_motion_detection_amcrest() and self._wanted_by(SWITCH):
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -306,7 +306,7 @@ class DahuaClient:
         # If we can't fetch, just assume 2 since that's pretty standard
         return 3
 
-    async def async_get_coaxial_control_io_status(self) -> dict:
+    async def async_get_coaxial_control_io_status(self, channel: int = 1) -> dict:
         """
         async_get_coaxial_control_io_status returns the the current state of the speaker and white light.
         Note that the "white light" here seems to also work for cameras that have the red/blue flashing alarm light
@@ -317,7 +317,7 @@ class DahuaClient:
         status.status.Speaker=Off
         status.status.WhiteLight=Off
         """
-        url = "/cgi-bin/coaxialControlIO.cgi?action=getStatus&channel=1"
+        url = "/cgi-bin/coaxialControlIO.cgi?action=getStatus&channel={channel}".format(channel=channel)
         return await self.get(url)
 
     async def async_get_lighting_v2(self) -> dict:
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -875,13 +875,15 @@ class DahuaClient:
         _LOGGER.debug("Setting coaxial control state to %s: %s", io, url)
         return await self.get(url)
 
-    async def async_set_nvr_coaxial_control_state(self, channel: int, enabled: bool) -> dict:
-        """Set an NVR-connected camera's active deterrence light state."""
+    async def async_set_nvr_coaxial_control_state(
+        self, channel: int, dahua_type: int, enabled: bool
+    ) -> dict:
+        """Set an NVR-connected camera's coaxial deterrence state."""
         io = 1 if enabled else 2
         url = (
             "/cgi-bin/coaxialControlIO.cgi?action=control&channel={channel}"
-            "&info[0].Type=1&info[0].IO={io}&info[0].TriggerMode=2"
-        ).format(channel=channel, io=io)
+            "&info[0].Type={dahua_type}&info[0].IO={io}&info[0].TriggerMode=2"
+        ).format(channel=channel, dahua_type=dahua_type, io=io)
         _LOGGER.debug("Setting NVR coaxial control state to %s: %s", io, url)
         return await self.get(url)
 
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -875,6 +875,16 @@ class DahuaClient:
         _LOGGER.debug("Setting coaxial control state to %s: %s", io, url)
         return await self.get(url)
 
+    async def async_set_nvr_coaxial_control_state(self, channel: int, enabled: bool) -> dict:
+        """Set an NVR-connected camera's active deterrence light state."""
+        io = 1 if enabled else 2
+        url = (
+            "/cgi-bin/coaxialControlIO.cgi?action=control&channel={channel}"
+            "&info[0].Type=1&info[0].IO={io}&info[0].TriggerMode=2"
+        ).format(channel=channel, io=io)
+        _LOGGER.debug("Setting NVR coaxial control state to %s: %s", io, url)
+        return await self.get(url)
+
     async def async_set_disarming_linkage(self, channel: int, enabled: bool) -> dict:
         """
         async_set_disarming_linkage will set the camera's disarming linkage (Event -> Disarming in the UI)
--- a/custom_components/dahua/config_flow.py
+++ b/custom_components/dahua/config_flow.py
@@ -26,6 +26,7 @@ from .const import (
     CONF_AUTO_DETECT_CHANNEL,
     CONF_USE_HTTPS,
     CONF_SCAN_INTERVAL,
+    CONF_NVR_ACTIVE_DETERRENCE,
     DEFAULT_SCAN_INTERVAL,
     MIN_SCAN_INTERVAL,
 )
@@ -324,6 +325,12 @@ class DahuaOptionsFlowHandler(config_entries.OptionsFlow):
                 default=self.options.get(CONF_SCAN_INTERVAL, DEFAULT_SCAN_INTERVAL),
             )
         ] = vol.All(vol.Coerce(int), vol.Range(min=MIN_SCAN_INTERVAL))
+        schema[
+            vol.Required(
+                CONF_NVR_ACTIVE_DETERRENCE,
+                default=self.options.get(CONF_NVR_ACTIVE_DETERRENCE, False),
+            )
+        ] = bool
         return self.async_show_form(
             step_id="user",
             data_schema=vol.Schema(schema),
--- a/custom_components/dahua/const.py
+++ b/custom_components/dahua/const.py
@@ -48,6 +48,7 @@ CONF_CHANNEL = "channel"
 CONF_AUTO_DETECT_CHANNEL = "auto_detect_channel"
 CONF_USE_HTTPS = "use_https"
 CONF_SCAN_INTERVAL = "scan_interval"
+CONF_NVR_ACTIVE_DETERRENCE = "nvr_active_deterrence"
 
 # Defaults
 DEFAULT_NAME = "Dahua"
--- a/custom_components/dahua/light.py
+++ b/custom_components/dahua/light.py
@@ -30,9 +30,12 @@ async def async_setup_entry(hass: HomeAssistant, entry, async_add_entities):
     if coordinator.is_flood_light():
         entities.append(FloodLight(coordinator, entry, "Flood Light"))
 
-    if coordinator.supports_security_light() and not coordinator.is_amcrest_doorbell():
+    if (
+        coordinator.supports_security_light() or coordinator.is_nvr_channel()
+    ) and not coordinator.is_amcrest_doorbell():
         #  The Amcrest doorbell works a little different and is added in select.py
-        entities.append(DahuaSecurityLight(coordinator, entry, "Security Light"))
+        security_light_name = "Warning Light" if coordinator.is_nvr_channel() else "Security Light"
+        entities.append(DahuaSecurityLight(coordinator, entry, security_light_name))
 
     if coordinator.is_amcrest_doorbell():
         entities.append(AmcrestRingLight(coordinator, entry, "Ring Light"))
@@ -168,7 +171,7 @@ class DahuaIlluminator(DahuaBaseEntity, LightEntity):
         profile_mode = self._coordinator.get_profile_mode()
         if self._coordinator.is_nvr_channel():
             await self._coordinator.client.async_set_nvr_coaxial_control_state(
-                self._coordinator.get_channel_number(), True
+                self._coordinator.get_channel_number(), SECURITY_LIGHT_TYPE, True
             )
         else:
             await self._coordinator.client.async_set_lighting_v2(channel, True, dahua_brightness, profile_mode)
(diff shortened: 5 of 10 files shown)


rroller (approved):

rroller:
@zios0007 I don't have an NVR. Does this PR look safe to you?

rroller:
Looks like the tests fail
```
Error: test_siren_asks_for_the_siren_not_the_security_light

AttributeError: '_Coordinator' object has no attribute 'is_nvr_channel'
Error: test_siren_turn_off_keeps_the_siren_type

AttributeError: '_Coordinator' object has no attribute 'is_nvr_channel'
Error: test_every_switch_is_named_after_the_device

AttributeError: 'DahuaSirenBinarySwitch' object has no attribute '_name'. Did you mean: 'name'?
...............................................FF...F...............  [137/137]
=================================== FAILURES ===================================
_____________ test_siren_asks_for_the_siren_not_the_security_light _____________
[gw1] linux -- Python 3.13.15 /opt/hostedtoolcache/Python/3.13.15/x64/bin/python3

    async def test_siren_asks_for_the_siren_not_the_security_light():
        """Both ride the same CGI; the type is the only thing separating them."""
        c = _Coordinator(channel=4)
        s = _switch(DahuaSirenBinarySwitch, c)
    
>       await s.async_turn_on()

tests/dahua/test_switch.py:106: 
```

zios0007:
I have an NVR, so I ran this against it. Short answer: **not safe as it stands**, but the direction is right and the gap is small. Two blockers, then detail.

### 1. CI is red, and deterministically so

`test_every_switch_is_named_after_the_device` fails because `tests/dahua/test_switch.py` builds entities with `object.__new__(cls)` and never calls `__init__`, so an instance attribute assigned only in `__init__` doesn't exist. Making it a class attribute fixes it and keeps the per-instance override working:

```python
class DahuaSirenBinarySwitch(DahuaBaseEntity, SwitchEntity):
    _name = "Siren"
```

The two `is_nvr_channel` failures are the fake coordinator not having the method — but see below, because I don't think the entity code should be calling it at all.

### 2. It creates deterrence entities on every NVR channel, capable or not

`supports_siren() or is_nvr_channel()` and `supports_security_light() or is_nvr_channel()` mean every channel on a recorder gets a Warning Light and an Alarm. On my 10-channel NVR that's 20 entities, most of them controlling nothing.

**This PR already contains the mechanism to avoid that, and then doesn't use it.** The probe fix in `__init__.py` — sending the real channel to `coaxialControlIO.cgi?action=getStatus` instead of the hardcoded `channel=1` — turns `_supports_coaxial_control` into an accurate per-channel signal. Measured just now on a DHI-NVR5464-16P-EI, one digest request per channel:

```
ch 1,2,3,4,5,8,12          ->  200   status.Speaker=Off  status.WhiteLight=Off
ch 6,7,9,10,11,13,14,15,16 ->  400   Error Bad Request!
```

All sixteen probed, and per `RemoteDevice` all are populated and enabled bar channel 16 — so this isn't "answered channels have a camera". An NVR does **not** uniformly mask this. `client.get` calls `raise_for_status()`, so those 400s surface as `ClientResponseError`, which is precisely what the existing `_supports_coaxial_control` probe already catches. So gate the entities on that instead:

```python
if coordinator.supports_siren() or coordinator.supports_coaxial_control():
```

(worth an accessor rather than reaching into `_supports_coaxial_control`, to match `supports_siren()` / `supports_security_light()` beside it.)

That gives you the outcome you're after — deterrence entities appear for NVR-attached TiOC cameras — without the phantoms on the other channels, and it removes the need for `is_nvr_channel()` in entity creation entirely.

I checked the ordering, since this only works if the probe has run by setup time: `async_config_entry_first_refresh()` is awaited at `__init__.py:113`, before `async_forward_entry_setups` at line 125. So the flag is populated when the platforms are built.

Worth noting *why* the entities were missing before, because it isn't the probe: `supports_security_light()` and `supports_siren()` are model-string tests, and on an NVR channel the model resolves to the recorder's (`DHI-NVR5464-16P-EI` here), so they were never going to fire. The probe result simply wasn't consulted for entity creation.

**`action=control` gives no signal at all, which is what makes the getStatus gate matter.** I fired `action=control&info[0].Type=1` at several channels, with and without `TriggerMode=2`, and got `200 OK` every time — *including on channel 6, whose getStatus returns 400*. There's no error path here. So a force-created entity on a camera with no deterrence hardware doesn't fail loudly: the switch flips in the UI, the CGI says OK, and nothing happens. That's the failure mode I'd least want to ship to every NVR user.

Two honest limits on my testing:

- I established that the probe discriminates *per channel*, which is enough to reject `is_nvr_channel()` as the gate. I could **not** establish what it discriminates *on* — channels 1-5, 8 and 12 answer while 6, 7, 9 and 10 don't, and those are all enabled `ProtocolType=Private` cameras on the same PoE subnet — the split doesn't track `Enable`, `ProtocolType` or `AudioInputChannels`, and `RemoteDevice` carries no model field to correlate against. Deterrence hardware is the likely explanation, not a demonstrated one. (The 400s do include everything you'd expect to fail: the ONVIF-attached device on 11, the two cameras off the PoE subnet on 13-14, the VTO doorbell on 15, and the disabled channel 16.)
- I could not verify `TriggerMode=2` either way. Both forms return an identical `200 OK` and neither moved `status.WhiteLight` on any channel I tried, so none of my cameras appear to have the hardware. I'm not disputing your result on the HFW3449T1-AS-PV — I just can't corroborate it, and nobody should read my testing as having done so.

A caveat on the gate itself: a 200 returns *both* `Speaker` and `WhiteLight`, so one probe gating both the Alarm and the Warning Light will still over-create whichever of the two a given camera lacks. Much better than twenty phantoms, but not exact.

### 3. `is_nvr_channel()` is unreliable at the edges

```python
return self._channel > 0 or "NVR" in self.model.upper()
```

`_channel` is the 0-based index, so channel 1 of a recorder is index 0 and falls through to the model test. That works on my NVR, but an **XVR/DVR or a rebadged recorder has no "NVR" in its model** — its channel 1 would take the direct-camera path while channels 2+ take the NVR path. Split behaviour on one device. Conversely any multi-channel non-NVR device with index > 0 gets misclassified.

**This mostly dissolves rather than needing to be fixed.** The PR uses the predicate in three places: entity creation, the probe's channel, and the `TriggerMode` decision. If entity creation gates on the probe (above), and the probe unconditionally uses `get_channel_number()` — which is what it should send regardless of device type — then `is_nvr_channel()` survives only to choose `TriggerMode=2`, where a wrong answer is far less costly than a phantom entity.

### Smaller things

- **The IO mapping isn't a change.** The PR lists "Map IO=1 to ON and IO=2 to OFF", but `async_set_coaxial_control_state` already does exactly that. Diffing the two methods, the only real differences are `&info[0].TriggerMode=2` and that the new one is called with `get_channel_number()` where the old one is called with `get_channel()`.

- **That channel argument is the real fix buried in here, and it deserves to be its own change.** Existing call sites pass `get_channel()` — the 0-based index — to a CGI whose own getStatus URL was hardcoded to `channel=1`, i.e. 1-based. Same bug as #600 on `ptz.cgi`. Sending the channel *number* is very likely correct on direct cameras too. I can't say the same for `TriggerMode=2` — I have an NVR and a VTO, no standalone deterrence camera — so keeping that conditional, as this PR does, is the right call.

- **The rename is safe.** `unique_id` is `serial + "_siren"`, a literal, not derived from the name — so Siren→Alarm changes only the friendly name and orphans nothing. It will rename an entity for NVR users who already had a working Security Light, which is cosmetic churn worth a release note.

- **The `FloodLight` branches are unreachable for NVR users.** `FloodLight` is only created when `is_flood_light()` matches a model string (ASH26/L26N/L46N/…), which an NVR channel's model won't. Its `else:` path still passes `get_channel()` and is untouched. No harm, but it inflates the diff.

- **Load side effect.** `client.get` keys the shared-read cache on the URL, so `...getStatus&channel=1` was one read shared by all ten of my entries; varying the channel makes it ten reads per poll. That's unavoidable and correct. Gating on `_supports_coaxial_control` wins most of it back, since `if self._supports_coaxial_control:` guards the poll and the 400 channels stop polling altogether.

- **The new test only covers the URL builder.** Nothing covers `is_nvr_channel`, entity creation, or the light/switch branching — which is exactly where CI broke and where the risk sits.

### What I'd ask for

1. Gate entity creation on the coaxial probe rather than `is_nvr_channel()`.
2. Send `get_channel_number()` to the probe unconditionally, not just for NVRs.
3. Make `_name` a class attribute so the existing tests pass.
4. A test for entity creation on a channel that reports no coaxial support.

With those, I'm happy to run it against my NVR and report back on real hardware.

zios0007:
One clarification on the `_name` fix above — my snippet was ambiguous, and taken literally it would break things. Keep the `__init__` parameter and *add* the class attribute as the fallback:

```python
class DahuaSirenBinarySwitch(DahuaBaseEntity, SwitchEntity):
    _name = "Siren"          # so instances built without __init__ still resolve name

    def __init__(self, coordinator, entry, name="Siren"):
        super().__init__(coordinator, entry)
        self._name = name
```

The class attribute is what makes `test_every_switch_is_named_after_the_device` pass, since that test builds entities with `object.__new__` and never calls `__init__`. Dropping the `__init__` in favour of the class attribute alone would make the existing `DahuaSirenBinarySwitch(coordinator, entry, name="Alarm")` call raise `TypeError`.

zios0007:
**Correcting my own recommendation above — the coaxial probe is not the right gate, and I have a better one.**

The owner of the NVR I tested on pointed out that one of the channels I'd written off actually has a white light. Chasing that down showed my suggested gate is far less precise than I claimed.

`Lighting_V2` reports, per channel, which physical lights a camera has. Fetched from my NVR, the **whole table**, not per channel:

```
table.Lighting_V2[1][0][0].LightType  = InfraredLight
table.Lighting_V2[1][0][1].LightType  = WhiteLight     <- channel 2
table.Lighting_V2[11][0][0].LightType = InfraredLight
table.Lighting_V2[11][0][1].LightType = WhiteLight     <- channel 12
table.Lighting_V2[11][0][2].LightType = AIMixLight
```

Exactly two channels have a white light. Set against the coaxial probe:

```
ch            1    2    3    4    5    6    7    8    9   10   11   12   13-16
coaxial     200  200  200  200  200  400  400  200  400  400  400  200   400
WhiteLight    -  yes    -    -    -    -    -    -    -    -    -  yes     -
```

So `_supports_coaxial_control` gives **seven** positives where only **two** cameras have the hardware. Gating on it is better than `is_nvr_channel()` — seven phantoms instead of ten — but it is not the clean fix I presented it as, and I'd rather flag that than let it get merged on my say-so.

### The gate I should have suggested

```python
def supports_white_light(self) -> bool:
    return self.data.get(f"table.Lighting_V2[{self._channel}][0][1].LightType") == "WhiteLight"
```

Three things make this the better signal:

- **It costs nothing.** `async_get_lighting_v2()` already fetches the whole `Lighting_V2` table every poll and `self.data` already holds these keys. No new request, no new probe, and it's a `getConfig` read so it rides the long cache TTL.
- **It's exact on the hardware I can see** — two positives, two cameras with lights.
- **Sub-index `[1]` is already the codebase's convention for the white light.** `FloodLight` and `DahuaSecurityLight` both read `Lighting_V2[channel][profile][1].Mode` today. This just asks whether that entry exists before building an entity around it.

It also drops `is_nvr_channel()` out of entity creation entirely, without needing the NVR/direct distinction at all — the same check works on a direct camera.

**One thing it does not cover:** this identifies the white light, not the siren. `Lighting_V2` says nothing about `Speaker`, so the Alarm switch still needs its own answer, and I don't have one. `supports_siren()`'s model-string test is the status quo there, and on an NVR channel it will keep returning False.

### On `TriggerMode=2`, I still can't corroborate it — and it now looks less certain

I ran `action=control&info[0].Type=1`, with and without `TriggerMode=2`, against both channels that genuinely have a white light. Every call returned `200 OK`, and *nothing observable changed* — not `coaxialControlIO` `getStatus`, not the `Lighting_V2` mode, on either channel, either form.

That isn't evidence against your patch: my NVR is a DHI-NVR5464-16P-EI on different firmware from your NVR4116HS, and I can't see the cameras to know whether the light physically lit. But it does mean the `200 OK` tells you nothing at all — the device returns it for channels whose getStatus 400s. If you have a way to confirm the state flips back in HA after a control call on your setup, that would be worth adding to the PR description, because on mine the status never moves and a user would see a switch that bounces straight back.

zios0007:
**Retracting the `Lighting_V2` gate I suggested above — it's wrong too. Please don't apply it.**

I got eyes on the hardware. Channel 2 advertises `table.Lighting_V2[1][0][1].LightType=WhiteLight`, and **it has no white light**. Verified rather than inferred:

- Forced the camera into night mode (`VideoInOptions[1].DayNightColor=2`) so a daylight lockout couldn't mask the result — my earlier test ran at 08:46 local and was worthless for that reason.
- Set `Mode=Manual` on the white-light entry across **all nine** day/night profiles, not just profile 0, so whichever was active got it. Held 25s.
- Then fired `coaxialControlIO action=control Type=1 IO=1 TriggerMode=2`. Held 25s.
- The NVR's owner watched the camera through both windows. Nothing lit. Config restored afterwards.

So `LightType=WhiteLight` in `Lighting_V2` is not evidence of a physical lamp, and my "exact on the hardware I can see" claim was wrong — I had two positives and had confirmed neither.

### Where that leaves capability detection on this NVR

Three signals, none of them trustworthy on a DHI-NVR5464-16P-EI:

| signal | ch2 (no light, confirmed) | ch12 (has a light) | verdict |
|---|---|---|---|
| `is_nvr_channel()` | true | true | fires on all 16 channels |
| `coaxialControlIO getStatus` | 200 | 200 | 200 on 7 of 16; 2 have lights at most |
| `Lighting_V2 LightType=WhiteLight` | **yes — but no lamp** | yes | disproven above |

The only config difference I can find between the two is that ch12 also lists `AIMixLight` and ch2 doesn't. That's n=2 and I'm explicitly *not* proposing it as a gate — I've now suggested two and been wrong twice.

### The part that worries me more

`action=control` has never produced an observable effect on this NVR — not on a channel with a light, not on one without, with or without `TriggerMode=2`, under forced night mode, checked against both `getStatus` and the `Lighting_V2` mode, and confirmed by eye. It returns `200 OK` every time, including for channels whose `getStatus` returns 400.

Worth noting the integration only speaks the HTTP CGI API, while this NVR reaches its cameras over `ProtocolType=Private` on port 37777. A `200 OK` from the NVR means the NVR accepted the HTTP request — not that anything crossed that link to the camera. That would explain everything I'm seeing, and it may well differ between your NVR4116HS and this 5464.

**None of this contradicts your testing** — you saw it work on your hardware and I believe you. But it does mean the behaviour isn't uniform across NVRs, which is exactly the risk in creating these entities for every NVR channel by default. On this NVR, every one of those 20 entities would be inert while reporting success.

I'd still suggest gating on *something* rather than `is_nvr_channel()`, and I'm happy to test any candidate signal against real hardware — I have a 16-channel NVR with a known-good channel and a known-bad one, which turns out to be a useful pair. I just can't offer you a signal I've verified myself yet.

zios0007:
**Final result from my hardware — this supersedes my back-and-forth above, and it changes my recommendation.**

I found a channel with a **confirmed physical white light** (channel 12, verified by the NVR's owner, who knows the camera and can drive its light from the Dahua app). It is the best possible test case for this PR. Every control path failed.

Forced the camera into night mode first, so no daylight lockout could mask it. Then, with the owner watching the camera through three separate 25-second windows:

| attempt | result |
|---|---|
| `Lighting_V2[11][*][1].Mode=Manual` (WhiteLight, all 9 profiles) | `200 OK`, no light |
| `Lighting_V2[11][*][2].Mode=Manual` (AIMixLight, all 9 profiles) | `200 OK`, no light |
| `coaxialControlIO action=control Type=1 IO=1 TriggerMode=2` | `200 OK`, no light |

`getStatus` stayed `WhiteLight=Off` throughout. Config restored and verified afterwards.

So on a **DHI-NVR5464-16P-EI**, the HTTP CGI deterrence path does not reach the camera at all — not this PR's route, not the route the existing security light uses. I couldn't isolate whether the NVR drops it or the camera refuses, because the cameras sit on an isolated PoE subnet (`10.2.2.0/24`) that isn't routable from outside the NVR.

### Why this matters for the PR

I'm not doubting your NVR4116HS result. Taken together they say the relay behaviour **differs by NVR model** — and that's the problem, because nothing in the API lets you tell which kind you're on:

- `action=control` returns `200 OK` unconditionally — including for channels whose `getStatus` returns `400`. It cannot report failure.
- `getStatus` returns `200` on my channel 12, the one with a real light that cannot be driven. So it doesn't separate "works" from "doesn't".
- `Lighting_V2 LightType` claims `WhiteLight` on a channel I've confirmed has no lamp at all.

All three signals are `true` for a channel where the feature is completely inert. **There is no probe that fails closed here.**

### What I'd suggest instead

Don't create these entities automatically for NVR channels. On my NVR this PR produces 20 entities that report success and do nothing — for a user, that's worse than the current state of having none, because a switch that silently lies is harder to diagnose than a missing feature.

Make it **opt-in**: a per-entry option like the existing `events` / `auto_detect_channel` toggles, defaulting off, so users who know their NVR relays coaxial control get the entities and everyone else is unaffected. That keeps your NVR4116HS working exactly as you've demonstrated, costs nothing for users on hardware where it doesn't, and needs no capability probe — which is just as well, because I couldn't find one that works.

The channel-number fix in this PR is worth keeping regardless, and would stand on its own as a small separate change.

Happy to test any candidate against this NVR — channel 2 (no lamp) and channel 12 (lamp, undriveable) make a useful pair.

Saufziege:
Thanks for the detailed testing and clarification.

I agree that automatic capability detection through `is_nvr_channel()`, `getStatus`, or `Lighting_V2.LightType` is not reliable across NVR models and firmware versions.

I will revise the PR as follows:

- Make NVR active-deterrence entities opt-in per config entry.
- Keep the regular Illuminator separate from Warning Light.
- Keep `Type=1` for Warning Light and `Type=2` for Alarm/Siren.
- Continue using the resolved NVR channel number for control and status requests.
- Fix the failing CI tests and add coverage for entity creation and channel/type handling.
- Document that NVR active-deterrence support is hardware- and firmware-dependent.

The default will remain disabled to avoid creating inactive entities on unsupported NVR channels.

zios0007:
**Context correction: I framed my findings as an NVR-relay problem. Your own issue tracker says the problem is broader than that.**

Three open issues describe exactly what I measured, on **direct cameras with no NVR involved**:

- **#555** (Mar 2026) — `coaxialControlIO.cgi` returns `Bad Request` on a DH-P5AE-PV. The reporter tried `TriggerMode` 0, 1 and 2 and `channel` 0, 1 and 2; every combination 400s. The siren works from DMSS, so the hardware is fine. They note it "works on some cameras but returns Bad Request on others."
- **#446** (Jan 2025) — security light and siren don't report state back; the toggle springs back to off while the light is physically on. Illuminator reports state correctly on the same camera.
- **#546** (Feb 2026) — TiOC 3 / TiOC Pro: light turns on, won't turn off, HA and DMSS disagree. TiOC 2 on the same network is fine.

So the two things I saw on my NVR — `400 Bad Request` on some channels, and `getStatus` never reflecting the light — are both **already-reported behaviour on direct cameras**. That makes "the NVR doesn't relay it" too narrow an explanation for what I measured, and I'd rather correct that than have it read as a clean NVR-vs-direct split.

It also cuts against the PR's second claim. #446 and #546 say state sync is broken on plain cameras, so `coaxialControlIO getStatus` isn't a reliable state source anywhere — which is consistent with my NVR never once reporting `WhiteLight=On`, including on a channel whose camera demonstrably has a working lamp.

**Relevant to how much certainty is available here:** Dahua's HTTP API spec is at v3.x now (v3.37 is in circulation), but the freely-available copies stop around v2.84, and 3.x needs a DEPP partner account and a signed NDA. More to the point, users comparing versions report that **v3.18 contains nothing TiOC-specific**. There's no documented NVR→camera passthrough for coaxial control either; the documented way to reach a camera behind an NVR is Virtual Host, which is a per-channel port mapping (`<nvr-ip>:6500N`) exposing the camera's *own* CGI, not an NVR API.

So every implementation of this is working from reverse-engineering, and the evidence in this repo says the endpoint behaves differently across models, firmwares, and TiOC generations.

That doesn't argue against your patch — it argues against turning it on by default. My suggestion stands: make it an opt-in per-entry option rather than automatic for every NVR channel. On hardware where the endpoint doesn't work, the current PR yields entities that report success and do nothing, and #446/#546 suggest that population is not small.

If it helps, #611 landing with an opt-in switch would also give #555, #546 and #446 a shared place to converge.

Saufziege:
Implemented the opt-in approach discussed above.

NVR active deterrence controls are now disabled by default and can be enabled per config entry via:

`Enable NVR active deterrence controls`

When enabled, NVR channels expose:

- Warning Light → `coaxialControlIO.cgi`, Type=1
- Alarm → `coaxialControlIO.cgi`, Type=2

Standalone camera behavior remains unchanged.

zios0007:
@Saufziege this addresses my objection properly — thank you for taking the opt-in route.

Default `False`, gated through `supports_nvr_active_deterrence()`, with the direct-camera path untouched, means an NVR user whose hardware ignores these commands gets nothing new rather than twenty entities that report success and do nothing. That was the whole of my concern. You also applied the `_name` fix exactly as described — class attribute *and* the `__init__` parameter — so the existing switch tests should pass again.

The only thing blocking it now is that it's `DIRTY`. Four PRs landed underneath you and shipped in 0.9.91, and two of them touch the exact lines you're changing. Here's the map, so the rebase is mechanical rather than exploratory.

### What will conflict

**`__init__.py`, the coaxial status fetch.** #632 gated every poll request on the platform that reads its answer. That line is now:

```python
if self._supports_coaxial_control and self._wanted_by(LIGHT, SWITCH):
    coros.append(asyncio.ensure_future(self.client.async_get_coaxial_control_io_status()))
```

Keep both sides — your channel argument and the `_wanted_by` gate. Dropping the gate would undo the change that reduces polling for everyone; dropping your argument puts the probe back on channel 1.

**`__init__.py`, the coordinator methods.** #630 added `_back_off_poll_interval` and `_restore_poll_interval`, #632 added `_wanted_by`. Your `supports_nvr_active_deterrence()` and `is_nvr_channel()` sit with the other `supports_*` accessors and shouldn't fight, but the file has moved a long way.

**`switch.py`.** #633 removed the `ClientError` import and the setup-time disarming-linkage probe (it now reads the capability the coordinator already has). #634 added `EntityCategory.CONFIG` to four switch classes and a `homeassistant.const` import. Your siren changes are in the same file.

**`const.py`.** #634 added `SENSOR` to `PLATFORMS`. Your `CONF_NVR_ACTIVE_DETERRENCE` should sit clear of it.

`light.py` I never touched, so that half should apply cleanly.

### Two small things, both non-blocking

`is_nvr_channel()` is still `self._channel > 0 or "NVR" in self.model.upper()`. Channel 1 of a recorder is index 0, so on an XVR or a rebadged unit — no "NVR" in the model — channel 1 takes the direct-camera path while channels 2+ take the NVR path. With the feature opt-in the blast radius is small, but it's split behaviour on one device.

And the probe still picks its channel with the same predicate. Sending `get_channel_number()` unconditionally would be correct on direct cameras too, and would leave `is_nvr_channel()` deciding only `TriggerMode`.

### For the record on my hardware

Still worth knowing before anyone enables this: on a DHI-NVR5464-16P-EI I could not make these commands do anything — not on a channel with a confirmed working white light, under forced night mode, with the owner watching. Every call returned `200 OK`. I'm not disputing your NVR4116HS result; the behaviour clearly differs by model, which is exactly why default-off is the right call.

Saufziege:
Rebased onto the latest `rroller/main` and resolved the conflicts.
The coaxial status polling now keeps the `_wanted_by(LIGHT, SWITCH)` gate while using the actual NVR channel number.

rroller (approved):

Issue #603, opened 2026-09-01
Reconnect login burst exceeds camera’s max connections (PF8-PV)


DeveloperSzkodziu:
Summary: Two Dahua PF8-PV cameras periodically lose their connection to the dahua integration every few days. Other camera models on the same setup/NVR never show this behavior. Camera recording itself is never interrupted (confirmed via DMSS — no gap in footage), so this is clearly an integration-to-camera connection issue, not a camera or network problem.

Environment: custom_components/dahua via HACS (thecode/ha-dahua fork), two Dahua PF8-PV cameras with up-to-date firmware, added to Home Assistant individually rather than through the NVR. A phone app (DMSS) connects only to the NVR, not directly to these cameras, so it’s not a competing client. Max connections on the cameras has been raised to 20 (the maximum allowed), which reduced but did not fully eliminate the problem.

What we’re seeing: Periodically, the integration goes into a reconnect cycle where it fires off a burst of near-simultaneous requests to the camera (system info, coaxial control status, PTZ status, lighting, disarming linkage, event stream attach, etc.). Checking the camera’s own system log during one of these episodes, we found roughly 10-15 Login and Logout entries firing within the very same second, all originating from Home Assistant’s IP address. This burst of logins from the integration is what exhausts the camera’s connection limit — not a slow leak building up over time, but a spike caused by the reconnect logic itself.

Once the camera’s connection limit is hit this way, trying to log into the camera’s own web UI directly returns “Exceeded maximum number of connections” — so the integration’s own reconnect behavior ends up locking itself (and anyone else) out of the camera by IP.

At the same time, most of the entities for that camera become unavailable, and the camera entity itself visibly falls back to a plain RTSP stream (visible right next to the camera entity in the entity list) instead of the normal integration-provided camera entity with its full set of related sensors and switches.

The pattern over time is very consistent: after a full camera restart, everything works perfectly for several days — all entities available, normal polling — and then, seemingly out of nowhere, the exact same failure cascade starts again (burst of logins → connection limit exceeded → entities drop → RTSP fallback), requiring another camera restart to recover. This cycle is consistent across both PF8-PV units, while the other camera models on the same NVR have never shown this behavior.

We’ve ruled out network issues (cameras are wired, recording never gaps), credential problems, and duplicate connections (camera isn’t added twice through two different integrations or via the NVR).

If you need more information - more than happy to support.

DeveloperSzkodziu:
[...]

PR #615, opened 2026-09-05
Hold one event stream per host, not one per channel

zios0007:
Closes the gap `#603` describes ("Reconnect login burst exceeds camera's max connections"), and does what the code's own comment has been asking for.

Attaching to a Dahua event stream returns **every** channel's events, regardless of which channel asked. So an NVR with eleven channels was holding eleven identical streams and having ten of them throw each event away. `on_receive` said as much:

> This is a short term fix. Right now for NVRs this integration creates a thread per channel to listen to events. Every thread gets the same response. We need to discard events not for this channel. **Longer term work should create only a single thread per channel.**

**Builds on #614** and shares its branch, so the diff currently shows that commit too. It collapses once #614 merges.

### The change

`DahuaHostEventStream`, keyed by address, holds one stream, attaches with the **union** of every channel's configured events, and hands each event only to the channels registered for its index.

```
before                          after
11 persistent connections       1
11 hourly reconnections         1
10 of 11 events discarded       none discarded
```

The union is load-bearing: attaching with a single channel's list would silently stop delivering codes another channel had selected. A test covers it, and narrowing it back to the owner's own list fails two tests.

### Three deliberate decisions

**A channel nobody has configured stays silent.** Every coordinator used to discard those events, so a dispatch table that helpfully started firing them would be a behaviour change dressed as a fix.

**Each channel gets its own copy of the event.** `handle_event` writes the device name into it, so sharing one dict between channels would have them scribbling on each other's.

**The stream moves off a departing owner.** It borrows a client, and unloading an entry closes that client's session. Reloading a single channel calls unload then setup on that entry alone — tearing the shared stream down there would have made the other ten deaf. That is the case `test_unloading_one_channel_leaves_the_others_streaming` exists for, and it is the mistake this design most invites.

### Verification

```
suite: 159 passed   (144 after #614)
```

Fifteen new tests, including one asserting eleven registered channels produce exactly **one** attach.

### Note

`normalize_address` here duplicates the helper added in #613. They are the same function; whichever lands second should drop its copy.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -342,6 +342,158 @@ async def _release_connector(address: str) -> None:
         await holder[0].close()
 
 
+class DahuaHostEventStream:
+    """One event stream for a host, shared by every channel configured on it.
+
+    The device's event stream is not per channel: attaching to it returns every
+    channel's events regardless of who asked. An NVR with eleven channels was
+    therefore holding eleven identical streams and having ten of them throw each
+    event away. This holds one, and hands each event to the channels that want
+    it.
+    """
+
+    def __init__(self, hass: HomeAssistant, address: str) -> None:
+        self._hass = hass
+        self._address = address
+        # channel index -> coordinators listening on that channel
+        self._by_channel: Dict[int, list] = {}
+        self._owner = None  # whose client the stream currently borrows
+        self._events: frozenset = frozenset()
+        self._task: asyncio.Task | None = None
+
+    @property
+    def coordinators(self) -> list:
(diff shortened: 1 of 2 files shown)


rroller (approved):

rroller:
This one has conflicts, feel free to merge

zios0007:
Rebased onto main now that #613 and #614 have merged. The conflict was the duplicated `normalize_address` this PR carried — #613's copy landed first, so as the commit message said, this one is dropped. Suite is 200 passed on the rebased branch.

PR #616, opened 2026-09-05
Share one read per host instead of one per channel

zios0007:
Addresses `#577` "massive login/logout in NVR log".

Eight of the nine CGI calls a coordinator makes every poll carry **no channel argument at all**:

```
MotionDetect  DisableLinkage  DisableEventNotify  SmartMotionDetect
Lighting_V2   VideoInMode     coaxialControlIO    ptz.cgi
```

An NVR with eleven config entries is therefore asked the same eight questions eleven times a cycle, and answers identically eleven times. Only `Lighting[channel][mode]` differs between them.

### The key is the URL

That is the whole design, and it is why this is safe: nothing has to be classified by hand as host-wide or per-channel. A read that carries a channel already has a different URL per channel, so `Lighting[0][0]` and `Lighting[1][0]` separate themselves and can never be confused for one another. There is no list to maintain and no list to get wrong.

It also means that when the hardcoded `coaxialControlIO.cgi?action=getStatus&channel=1` is fixed to use the real channel, the URLs stop matching and the sharing on that URL stops with them — correctly, with no change here.

### What it does

- Two entries asking the same question at the same moment cost **one** round trip between them. Registration happens before the request starts, so callers queued on the per-host limiter from #606 share the slot rather than each waiting for their own.
- A repeat within a five second TTL costs none.
- Only reads are cached. Anything else is a write, and a write drops the host's entries — a write is how these values change.

I am not going to quote a request-count figure. HA restarts each coordinator's interval from its own last completion, so eleven entries drift apart by however long their setups took, and I have not measured the real alignment on a live NVR. What is certain: reads that overlap now cost one request, repeats within the TTL cost none, and the reduction scales with alignment up to about 5x in the aligned case.

### Three things it has to get right

**One entry's cancellation must not take the read from the other ten.** The shared read is a task every caller awaits under `asyncio.shield`. A reload or a timeout on one channel is not the other ten channels' problem — and #613's HTTPS repair flow reloads every entry on a host in sequence, so this is reachable.

**A failed read is not remembered.** The expiry is stamped only once the request lands, so a device that just came back is asked again rather than told no for the rest of the TTL.

**A write during an in-flight read wins.** The write drops the entry; the in-flight read settles for whoever is already waiting and does not put itself back. Otherwise a toggled switch would spring back for five seconds.

Callers get a copy, not the cached dict.

### Verification

```
suite: 342 passed   (324 before)
```

Eighteen new tests. Each property was proven by reintroducing the bug and confirming the right test fails:

| broken | tests that fail |
|---|---|
| coalescing removed | 6 |
| credentials dropped from the key | `test_two_users_on_one_host_do_not_share_a_read` |
| writes no longer invalidate | 3 |
| a failed read treated as usable | `test_a_failed_read_is_not_remembered` |
| `shield` removed | `test_one_cancelled_entry_does_not_take_the_read_away` |
| result returned by reference | `test_callers_cannot_scribble_on_each_others_result` |
| in-flight read republished after a write | `test_a_write_while_a_read_is_in_flight_does_not_republish_it` |

The first pass had a redundant failure-cleanup branch that no test could distinguish; it is gone rather than papered over with a test written to fit it.

Rebased onto main now that #615, #617, #618 and #619 have merged.

### Landing after #617 changed two things worth calling out

**The digest tests were quietly hollowed out, and are fixed here.** `test_shared_digest_state.py` had eleven clients read the *same* URL. With reads now shared by URL, those eleven collapse into one request — so the tests stopped observing the digest state at all. `test_the_count_keeps_rising_across_entries` was left with a single nonce count to check for monotonicity, and passed for that reason. Each client now reads a different URL, which is also what a real channel does: nine different reads a poll, one shared challenge across all of them. Reverting to per-client digest state still fails the same seven tests it failed on #617, so they still bite.

**The autouse fixture went in the wrong file.** It was a second `conftest.py` at the repo root, beside the real `tests/conftest.py`. It is now in `tests/conftest.py` where it belongs.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -24,7 +24,7 @@ from homeassistant.helpers.typing import ConfigType
 from homeassistant.const import EVENT_HOMEASSISTANT_STOP
 
 from . import dahua_utils
-from .client import DahuaClient
+from .client import DahuaClient, clear_host_cache
 from .model_profiles import is_sdt4e425
 
 from .const import (
@@ -339,6 +339,7 @@ async def _release_connector(address: str) -> None:
     holder[1] -= 1
     if holder[1] <= 0:
         _HOST_CONNECTORS.pop(address, None)
+        clear_host_cache(address)
         await holder[0].close()
 
 
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -2,6 +2,7 @@
 import logging
 import socket
 import asyncio
+import time
 import aiohttp
 import async_timeout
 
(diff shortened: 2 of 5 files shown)


rroller:
Has conflicts

rroller:
has more conflicts

rroller (approved):

PR #621, opened 2026-09-05
Point smart motion detection at the channel it belongs to

zios0007:
Same family as #619, but confirmed against real hardware rather than reasoned from the response shape — and unlike #619 it is wrong on **both** sides.

```
read:  table.SmartMotionDetect[0].Enable        __init__.py
write: setConfig&SmartMotionDetect[0].Enable    client.py
```

`SmartMotionDetect` is indexed by channel, exactly like `MotionDetect` — which the coordinator already reads correctly with `table.MotionDetect[{channel}].Enable`. Only this one was left at `[0]`.

### What the device actually returns

`getConfig&name=SmartMotionDetect` on a `DHI-NVR5464-16P-EI` with ten channels configured:

```
table.SmartMotionDetect[1].Enable=true
table.SmartMotionDetect[1].ObjectTypes.Human=true
table.SmartMotionDetect[1].ObjectTypes.Vehicle=true
table.SmartMotionDetect[1].Sensitivity=Middle
table.SmartMotionDetect[11].Enable=true
...
```

**The rows are sparse** — only the channels the option is configured on — and **there is no row 0**. So on that NVR:

- every one of the ten switches read `false`, permanently, including channel 1 which is genuinely `true`
- toggling any of them wrote channel one's option, not the camera the switch belonged to

`MotionDetect` on the same device returns rows `0-9, 11, 14`, and `VideoInMode` returns `0-9, 11`, so sparse indices are normal here and not particular to this table.

### Single cameras are untouched

One row is all a single camera reports, and that row is row 0. A camera on channel 0 produces a byte-identical URL on the write and reads the same key it always did:

```
http://host:80/cgi-bin/configManager.cgi?action=setConfig&SmartMotionDetect[0].Enable=false
```

That is asserted directly, not inferred.

### Verification

```
suite: 341 passed   (324 before)
```

| broken | tests that fail |
|---|---|
| read back to row 0 | 4 |
| write back to row 0 | 4 |
| single-camera fallback removed | `test_a_channel_with_no_row_of_its_own_falls_back_to_row_zero` |

The write row is there because of a gap this PR found in its own tests. On the first pass, reverting the write to `[0]` **passed everything** — `test_switch.py` asserts the arguments the switch hands to the client, and nothing asserted the URL the client then builds, so a hardcoded index was invisible at that layer. The new tests drive the real `DahuaClient` against a recording session and assert the URL. `test_switch.py` is updated to expect the channel argument.

Independent of #616 and #620.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -1129,11 +1129,21 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
         return self.data.get("table.DisableEventNotify.Enable", "").lower() == "false"
 
     def is_smart_motion_detection_enabled(self) -> bool:
-        """ Returns true if smart motion detection is enabled """
+        """ Returns true if smart motion detection is enabled
+
+        SmartMotionDetect is a host-wide read that returns a row per channel,
+        and the rows are sparse: a device only reports the channels the option
+        is configured on. Reading row 0 for every channel reported one camera's
+        setting for all of them, and reported false for the whole NVR when
+        there is no row 0 at all.
+        """
         if self.supports_smart_motion_detection_amcrest():
             return self.data.get("table.VideoAnalyseRule[0][0].Enable", "").lower() == "true"
-        else:
-            return self.data.get("table.SmartMotionDetect[0].Enable", "").lower() == "true"
+        value = self.data.get("table.SmartMotionDetect[{0}].Enable".format(self._channel))
+        if value is None:
+            # A single camera reports one row, and that row is row 0.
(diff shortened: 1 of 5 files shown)


zios0007:
The failing `Run tests` check here is not this PR. It is `test_stream_jitter.py::test_eleven_streams_land_in_different_seconds` — a flaky test I introduced in #614 that is already on main and fails 7.5% of all CI runs on the repository. The failure was `assert 10 == 11`: eleven jittered lifetimes colliding into ten distinct seconds, which the birthday problem says happens about one run in thirteen.

#622 fixes it. Re-running this check should pass; merging #622 first makes it stop happening on every open PR.

rroller (approved):

Issue #526, opened 2025-10-31
cancel_call stop working in VTO2211G-WP


DAVIZINH0:
Hello,
i have a VTO2211G-WP and after upgrade to this firmware:
2025-03-31 V4.700.0000000.3.R

the service "dahua.vto_cancel_call" stop working for me
do you know if is a general problem or my fault?

Feneck91:
Same. And no possibility to set VTO volume to 0 (mute).

altrdev:
I downgraded the firmware for using vto_cancel_call. It's a pity because 4.700 fixed audio codec issues, but removed the vto_cancel_call ability. Now I'm on 4.511 and use @felipecrs sh for fix vto audio codec

DAVIZINH0:
i want to downgrade, but my problem is my vto is VTO2211G-WP  and i cant find the firmware :-(
edit: i find one in the ftp that post @Feneck91 
i will try! thanks

DAVIZINH0:
> I have the VTO2211G-P-S2 and you have VTO2211G-WP, not really the same. I think Dahua is blocking the downgrade.

maybe dahua is blocking :-(

felipecrs:
Also try:

```
curl -fsSL --digest --globoff http://admin:password@192.168.1.40/cgi-bin/VideoTalkPeer.cgi?action=hangup
```

felipecrs:
Just press the doorbell button, then when it's ringing, try running the commands in your computer. We just want to check if they work.

Feneck91:
When I write http://<user>:<password>@192.168.0.65/cgi-bin/VideoTalkPeer.cgi?action=hangup
With action = cancel / hangup / refuse the reply is:
Error
Bad Request!
 
When I write http://<user>:<password>@192.168.0.65/cgi-bin/videotalkpeer.cgi?action=hangup
Error
Not Implemented!

The URL is case sensitive? Missing some things?
What Is VTT / VTA products?

M3c4tr0x:
Great job !

I looked at @Feneck91's pcapng file using a packet analyzer and filtered only the SIP packets. This is what I found:
<img width="1469" height="192" alt="Image" src="https://github.com/user-attachments/assets/0ccb0006-9baa-451c-9c0c-5e5cb3181013" />
I’m not sure if this will help, but it looks like the “cancel” is actually showing as "BUSY HERE". I’m not familiar with the SIP protocol at all.
Hopefully, this can be of some use to you.

[...]

PR #623, opened 2026-09-05
Add reboot and open door buttons

### LINKED ISSUE #552: Button entities
Button entities is somewhat modern in Home Assistant.

It would be great if this integration provided button entities for some things, like:

- Cancel VTO Call
- Open door
- Reboot

And probably even more like PTZ controls. Similar to the [official Reolink integration](https://www.home-assistant.io/integrations/reolink#button-entities), which is Home Assistant-certified.

Button entities are discovered much easier and they are also much simpler to invoke manually, or even through automations.

One benefit is that I could disable the camera entity, which I don't need anyway, and could instead of have the button entity. Currently I cannot disable the camera entity because the service/action depends on it.

zios0007:
Closes #552, which explicitly benchmarks this integration against the HA-certified Reolink one.

`button.py` has been on disk since 2021 containing an empty `async_setup_entry` and this:

> `# TODO: Add some buttons. This requires a pretty recent version of HomeAssistant so I'm waiting a bit longer before adding buttons`

That refers to **2021.12**. It was also never listed in `PLATFORMS`, so the file never loaded regardless.

### Two buttons

| Button | Call | Added when |
|---|---|---|
| Reboot | `magicBox.cgi?action=reboot` | always — `RESTART` device class, `CONFIG` category |
| Open Door | `accessControl.cgi?action=openDoor` | `coordinator.is_doorbell()` |

**Reboot does not refresh afterwards.** The device is on its way down; polling it there would turn a successful press into an error in the log.

**Open Door is gated on the doorbell check.** The endpoint is a VTO operation, and a button that always errors is worse than no button.

### Cancel Call is deliberately not here

It looks like the obvious third button `#552` asks for, and it isn't:

```python
async def cancel_call(self):        # vto.py
    def cancel(message): ...
    self.send("console.runCmd", cancel, {"command": "hc"})
    return True
```

- `async` with no `await` — it pushes a message onto the VTO socket and returns `True` unconditionally. Awaiting it confirms nothing.
- Different transport: the VTO protocol on port 5000, not the HTTP CGI client the other buttons use, through `_vto_client` which is `None` until that socket connects.
- **`#526` "cancel_call stop working in VTO2211G-WP" is the most-commented open bug in this repository.** Wrapping a known-broken call in a new entity would attract reports blaming the button.

It needs a real success signal from the VTO response and a `HomeAssistantError` rather than an `AttributeError` when there is no client, and that is a separate change.

### Two consequences worth flagging

Adding a platform adds a toggle to the options screen, which is built from `sorted(PLATFORMS)` — so `en.json` gains a `button` label, and a test asserts every platform has one rather than showing as a bare key.

The identity table from #618 gains `_reboot` and `_open_door`. It is the only place these strings exist as literals, so a new platform has to be recorded there; renaming either suffix now fails it in five places.

### Verification

```
suite: 346 passed   (324 before)
```

Renaming `_reboot` to `_restart` fails 5 identity-table assertions and 2 button tests.

One of the new tests was wrong on the first pass: it asserted `DahuaRebootButton._attr_device_class is ButtonDeviceClass.RESTART` on the class. Home Assistant rewrites class-level `_attr_` assignments into cached properties, so that read returns a descriptor, not the value. It now reads `device_class` off an instance, which is what HA actually uses.

Independent of #616, #620, #621 and #622.

### DIFF
--- a/custom_components/dahua/button.py
+++ b/custom_components/dahua/button.py
@@ -1,12 +1,73 @@
 """
 Button entity platform for Dahua.
 https://developers.home-assistant.io/docs/core/entity/button
-Buttons require HomeAssistant 2021.12 or greater
 """
+import logging
+
+from homeassistant.components.button import ButtonDeviceClass, ButtonEntity
+from homeassistant.const import EntityCategory
 from homeassistant.core import HomeAssistant
+
+from custom_components.dahua import DahuaDataUpdateCoordinator
+
+from .const import DOMAIN
+from .entity import DahuaBaseEntity
+
+_LOGGER = logging.getLogger(__package__)
 
 
 async def async_setup_entry(hass: HomeAssistant, entry, async_add_devices):
     """Setup the button platform."""
-    # TODO: Add some buttons. This requires a pretty recent version of HomeAssistant so I'm waiting a bit longer
-    # before adding buttons
+    coordinator: DahuaDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
+
+    buttons = [DahuaRebootButton(coordinator, entry)]
+
+    # Opening a door is a VTO operation. On anything else the endpoint is not
+    # there, and a button that always errors is worse than no button.
+    if coordinator.is_doorbell():
+        buttons.append(DahuaOpenDoorButton(coordinator, entry))
+
+    async_add_devices(buttons)
+
+
+class DahuaRebootButton(DahuaBaseEntity, ButtonEntity):
+    """Reboots the device."""
+
+    _attr_device_class = ButtonDeviceClass.RESTART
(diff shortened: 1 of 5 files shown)


zios0007:
The `Run tests` failure is not this PR — it is `test_stream_jitter.py::test_eleven_streams_land_in_different_seconds` on main, failing with `assert 10 == 11`. That is the flaky test #622 fixes: eleven jittered lifetimes colliding into ten distinct seconds, which happens on about 7.5% of all CI runs in this repository.

It has now hit two of these PRs in a row (#621 and this one) within an hour, which is roughly what a 1-in-13 failure rate looks like. Merging #622 first stops it happening on every open PR.

rroller (approved):

PR #624, opened 2026-09-05
Stop the camera claiming WebRTC it cannot serve

zios0007:
`camera.py` set this on every camera:

```python
self._attr_frontend_stream_type = StreamType.WEB_RTC
```

This integration serves RTSP through the stream component and has no `async_handle_async_webrtc_offer`. The claim says the opposite of what it does.

It is also **dead**. Home Assistant derives `camera_capabilities` from `_supports_native_async_webrtc` and `_webrtc_provider`; nothing reads `_attr_frontend_stream_type`:

```python
if CameraEntityFeature.STREAM in self.supported_features:
    if self._supports_native_async_webrtc:
        frontend_stream_types.add(StreamType.WEB_RTC)
    else:
        frontend_stream_types.add(StreamType.HLS)
        if self._webrtc_provider:
            frontend_stream_types.add(StreamType.WEB_RTC)
```

Asking the real capability logic both ways returns the same answer:

```
with the WEB_RTC line : CameraCapabilities(frontend_stream_types={<StreamType.HLS: 'hls'>})
without it            : CameraCapabilities(frontend_stream_types={<StreamType.HLS: 'hls'>})
```

So this changes no behaviour on any Home Assistant version the integration supports — it removes a line that misstates the transport, and the now-unused `StreamType` import with it.

The three new tests pin the answer rather than just asserting the deletion, so a future WebRTC claim has to arrive with an offer handler behind it.

```
suite: 327 passed   (324 before)
```

Independent of #616, #620, #621, #622 and #623.

### DIFF
--- a/custom_components/dahua/camera.py
+++ b/custom_components/dahua/camera.py
@@ -6,7 +6,7 @@ import voluptuous as vol
 
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import entity_platform
-from homeassistant.components.camera import Camera, CameraEntityFeature, StreamType
+from homeassistant.components.camera import Camera, CameraEntityFeature
 
 from custom_components.dahua import DahuaDataUpdateCoordinator
 from custom_components.dahua.entity import DahuaBaseEntity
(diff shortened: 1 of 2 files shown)


zios0007:
Force-pushed to add the three tests. The original push was missing them entirely — I used `git commit -a`, which does not stage untracked files, so the branch contained only the one-line `camera.py` change while the description claimed tests and a suite count that could not have come from it. The diff now matches what the description says.

rroller (approved):

PR #627, opened 2026-09-05
Report event stream failures instead of swallowing them

zios0007:
`stream_events` caught every exception, logged it at **DEBUG**, and returned normally. So the caller's warning —

```python
_LOGGER.warning("Event stream for %s ended unexpectedly: %s", ...)
```

— was unreachable for every network, HTTP, auth and timeout failure it was written to report.

That silence is not theoretical. Someone reporting constant device-side activity from their NVR had **literally nothing** in their Home Assistant log about it, which made the problem look like it wasn't the integration at all. It cost a lot of diagnosis time.

### Three ways the stream could end leaving no trace anywhere

| | before | now |
|---|---|---|
| any exception | DEBUG, inside `stream_events` | propagates; caller warns |
| **clean close** — device answers 200 and ends the response | raises *nothing*; not even the DEBUG line ran | `EventStreamClosed` |
| no credentials | returned instantly, caller then spun a silent 60s retry loop forever | `EventStreamClosed` |

The middle one is the worst: a device that refuses `action=attach` by accepting it and immediately closing produced **no output at any log level**, forever. A long poll that returns is a failure, not a result.

### Once per outage, not once per retry

Silence was one extreme; a warning every sixty seconds forever is the other. The host stream tracks whether it is already failing and logs the transition — WARNING on the first failure, DEBUG while it persists.

### The retry delay had a branch with no delay at all

```python
if elapsed < 10:  ->  sleep(60)
else:             ->  reconnect immediately
```

A device closing the socket at eleven seconds hit the `else` every time and reconnected as fast as it could, forever, silently. Now graduated:

```
< 10s   -> ~60s     (unchanged)
< 60s   -> ~10s     (new; this band had no delay)
>= 60s  -> immediate (a healthy stream being recycled must not pause)
```

Worst-case re-attach rate is now one per ten seconds instead of unbounded.

### Verification

```
suite: 334 passed   (324 before)
```

| broken | tests that fail |
|---|---|
| clean close goes back to silent | 3 |
| credentials guard removed | `test_missing_credentials_raise_instead_of_looping_silently` |
| back to the no-delay branch | 2 |

That middle row is there on the second attempt. The first version of that test only asserted `EventStreamClosed` was raised — which the *end-of-stream* raise also satisfies, so the guard could be deleted with every test still green. It now asserts the reason and that no request was made.

Two existing tests in `test_event_stream.py` are updated: they pinned the old swallow-everything contract. One of them needed `aiohttp.ServerTimeoutError` rather than `asyncio.TimeoutError`, since the read timeout subclasses the latter and was indistinguishable from the outer `wait_for`.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -62,6 +62,21 @@ EVENT_STREAM_JITTER = 0.1
 # would otherwise be retried by every channel at once, sixty seconds later.
 EVENT_STREAM_RETRY_SECONDS = 60
 
+# A stream that lived this long was working, so reconnect at once. Anything
+# shorter gets backed off, because the fast path used to have no delay at all:
+# a device closing the socket at eleven seconds reconnected forever, silently.
+EVENT_STREAM_HEALTHY_SECONDS = 60
+EVENT_STREAM_SHORT_RETRY_SECONDS = 10
+
+
+def event_stream_retry_delay(lived_seconds: float) -> float:
+    """How long to wait before re-attaching, given how long the stream lasted."""
+    if lived_seconds < 10:
+        return jittered(EVENT_STREAM_RETRY_SECONDS)
+    if lived_seconds < EVENT_STREAM_HEALTHY_SECONDS:
+        return jittered(EVENT_STREAM_SHORT_RETRY_SECONDS)
+    return 0.0
+
 
 def jittered(seconds: float, fraction: float = EVENT_STREAM_JITTER) -> float:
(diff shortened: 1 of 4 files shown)


rroller (approved):
This has conflicts

zios0007:
Rebased onto current main. The conflict was just the #616 host-cache block and this PR's `EventStreamClosed` landing in the same spot in `client.py` — both kept, no logic changed. Checked it doesn't reintroduce `async_timeout` after #629. CI green, mergeable.

PR #628, opened 2026-09-05
Stop re-reading settings that nothing has changed

zios0007:
**Stacked on #616** — that PR adds the shared read cache; this one gives it two lifetimes. The diff shows both until #616 merges.

Reads fall into two kinds and were treated as one.

**Status reads** change on their own — where a PTZ camera points, whether the siren is sounding, which day/night profile the device switched itself to. Those must be polled.

**Config reads** change only when something writes them. A write already drops this host's cached entries, so a change made through Home Assistant is reflected immediately. Re-asking the device every poll buys nothing but load.

```
getConfig  ->  300s   MotionDetect, DisableLinkage, DisableEventNotify,
                      SmartMotionDetect, Lighting, Lighting_V2, VideoInMode
everything ->    5s   ptz.cgi getStatus, coaxialControlIO getStatus
```

### Why this is worth more than one request per poll

Measured against a real NVR while diagnosing a report of once-a-second device log entries:

```
Connection: close                        on every response
call 1: [(401, 'none'), (200, 'AUTH')]
call 2: [(401, 'AUTH'), (200, 'AUTH')]   <- cached challenge refused
call 3: [(401, 'AUTH'), (200, 'AUTH')]
```

The device closes the connection after every response **and** reissues its digest nonce every time, so the cached challenge is always rejected — it sends no `stale="true"` and no `Authentication-Info: nextnonce`. **Each logical call therefore costs two TCP connections, two HTTP requests, and one refused login written to the device's own log.**

Seven of the ten calls in a poll cycle are settings reads. Removing them removes about two-thirds of that.

### This is the part that helps a single camera

Worth being explicit, because it is the gap in the recent work: #615, #616 and #617 all reduce load by sharing between config entries, which does **nothing** for someone with one camera. That is exactly the configuration in the report that prompted this. One camera shares nothing with anybody; what helps it is not asking twice.

### The trade

A change made in the Dahua app or web UI, rather than through Home Assistant, can take up to five minutes to appear. Changes made through Home Assistant are immediate, because the write invalidates the host's cache — there is a test for that.

### Verification

```
suite: 348 passed   (324 before)
```

| broken | tests that fail |
|---|---|
| one lifetime for everything again | 2 |
| status cached as long as settings | 2 |

Both behavioural tests advance a fake clock by a poll interval between cycles. Without that, five calls in a row fit inside any TTL and the test proves nothing — the first version passed against both mutations for exactly that reason.

### DIFF
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -71,8 +71,38 @@ def _digest_state(address: str, username: str) -> dict:
 # carry a channel has a different URL per channel and is never shared. Only
 # reads are cached; anything else drops the host's entries, because a write is
 # how these values change.
+
+# How long a shared read answers for. Two lifetimes, because the reads fall
+# into two kinds.
+#
+# Status reads change on their own: where a PTZ camera is pointing, whether the
+# siren is sounding, which day/night profile the device has switched itself to.
+# Those have to be polled, so they get a lifetime short enough to only cover
+# one poll's fan-out.
+#
+# Config reads change only when something writes them -- and a write drops this
+# host's entries, so a change made through Home Assistant is reflected at once.
+# Re-asking the device every poll buys nothing except load. On hardware
+# measured for this, each of those calls costs two TCP connections, two HTTP
+# requests and one refused login in the device's own log, because the device
+# answers Connection: close and reissues its digest nonce every time. A camera
+# with no dashboard open was spending most of its request budget re-reading
+# settings nobody had touched.
+#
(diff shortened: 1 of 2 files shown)


rroller (approved):
This has conflicts

zios0007:
Rebased onto current main. #616 has since merged, so this is now just the config-read TTL on its own — the stacked commit dropped out cleanly. Only fixed up blank-line spacing where the two comment blocks met. CI green, mergeable.

PR #514, opened 2025-10-11
fix: skip PTZ status probe on NVRs; gate unsupported platforms

NirBY:
Fixes #513

- Skip ptz.cgi?action=getStatus probe on NVR models (400/timeout)
- Add timeout + exception handling around PTZ probe
- Only load 'switch' when coaxial control supported; skip 'light' when no lighting

PR #632, opened 2026-09-07
Stop fetching answers no loaded platform will read

zios0007:
Every entry can already switch off `camera`, `switch`, `light`, `select` and `binary_sensor`. Those toggles gate `async_forward_entry_setups` — entity creation — and nothing else. The poll decides what to request purely from what the device reports supporting and never looks at them.

So an entry with `select` switched off still calls `ptz.cgi?action=getStatus` on every single cycle, to populate a preset position entity that was never created.

That isn't free. Measured on a Dahua NVR, every logical CGI call costs **two TCP connections, two HTTP requests and one refused login**, because the device answers `Connection: close` and reissues its digest nonce each time, so a cached challenge is always rejected. PTZ status and coaxial status are also the two per-poll reads that #628's config cache does not cover, so they are paid in full on every cycle.

### The change

Follow the toggles that already exist rather than adding a dozen new "poll this?" options. Each request is gated on the platform that reads its answer:

| request | gated on |
|---|---|
| `ptz.cgi getStatus` | `select` |
| `coaxialControlIO getStatus` | `light` **or** `switch` |
| `MotionDetect` | `camera` **or** `switch` |
| `Lighting_V2`, infrared, `LightGlobal` | `light` |
| `DisableLinkage`, `DisableEventNotify`, `SmartMotionDetect` | `switch` |

The mapping was derived by finding each accessor's callers rather than by assuming — `is_siren_on` in `switch.py`, `is_security_light_on` in `light.py`, and so on. The two requests with a second reader keep working when either platform alone is enabled, which is what the `or` cases pin. Binary sensors read the event stream rather than the poll, so they gate nothing.

**The profile mode read is deliberately left alone.** `diagnostics.py` reports it as well as the light platform, and it is a `getConfig`, so #628 already caches it — gating it would risk a stale diagnostics field for no real saving.

**Read from the entry options, not `coordinator.platforms`.** The first refresh runs before the platforms are forwarded, so that list is still empty at that point and gating on it would skip every read on the first poll.

### Tests

The failure mode here is silent — map a request to the wrong reader and an entity quietly shows a default instead of erroring — so the tests pin the mapping rather than the mechanism:

- every request is still fetched with all platforms on, so the default is unchanged
- each request disappears when the platform that reads it is switched off
- the two shared requests survive losing *either* reader, and stop only when both are off
- switching off one platform doesn't drop another platform's reads

### Where it came from

@DeveloperSzkodziu suggested this in #603, having noticed their PF8-PV generates far more traffic than their other cameras. It does: it is PTZ **and** active deterrence, so it qualifies for both of the uncached per-poll reads, while a fixed camera makes neither. They don't use PTZ in Home Assistant at all and are currently paying for it every cycle.

Related to #514, which approaches this from the capability side — don't load a platform at all when the device supports nothing on it. Worth noting that PR has been untouched since 2025-10-11 and no longer merges, and its other half (exception handling around the PTZ probe) is already in main. The part still undone is orthogonal to this change rather than overlapping: it decides which *platforms* to load, this decides which *requests* to make for the platforms that did load.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -35,6 +35,10 @@ from .const import (
     CONF_NAME,
     DOMAIN,
     PLATFORMS,
+    CAMERA,
+    LIGHT,
+    SELECT,
+    SWITCH,
     CONF_RTSP_PORT,
     STARTUP_MESSAGE,
     CONF_CHANNEL,
@@ -758,6 +762,20 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
             )
             self.update_interval = interval
 
+    def _wanted_by(self, *platforms: str) -> bool:
+        """Whether anything that reads this answer is actually loaded.
+
+        The poll used to fetch purely on what the device reported supporting,
+        so an entry with `select` switched off still paid for a PTZ position
+        read on every cycle to feed an entity that was never created. Each of
+        those costs the device a connection and a login it has to refuse.
+
+        Read from the entry options rather than `coordinator.platforms`: the
+        first refresh runs before the platforms are forwarded, so that list is
+        still empty then and everything would be skipped on the first poll.
+        """
+        return any(self.config_entry.options.get(platform, True) for platform in platforms)
+
     async def _async_update_data(self):
         """Reload the camera information"""
         data = {}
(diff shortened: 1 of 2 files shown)


rroller (approved):

rroller:
has merge conflicts

zios0007:
Rebased onto main now that #630 has landed — thanks for merging that one.

The conflict was just the two PRs adding methods at the same point in the coordinator: #630's `_back_off_poll_interval`/`_restore_poll_interval` and this one's `_wanted_by`. Both kept, no logic changed on either side. CI green and mergeable.

They complement each other rather than overlapping: #630 reduces requests to a device that has stopped answering, this one reduces them to a device that is answering fine but whose answers nobody is reading.

rroller (approved):

PR #638, opened 2026-09-08
Read the day/night profile from whichever field the camera actually uses

### LINKED ISSUE #605: 0.9.84 regression: illuminator control writes inactive profile on cameras in 'General' profile mode (Config[0]=2)
## Regression in 0.9.84: white-light control broken on cameras using the "General" profile (Config[0]=2)

Since updating 0.9.83 → 0.9.85, turning on the illuminator entity no longer physically lights the camera's white light, while HA shows the entity as `on`. Bisected to the profile-detection change from #582 ("Fixed day/night profile control on dual illuminator cameras").

**Camera:** EmpireTech (Dahua OEM) IP camera, profile management set to **General** (single profile for all conditions).

**What the camera reports:**
```
table.VideoInMode[0].Config[0]=2      # active profile: 2 = General
table.VideoInMode[0].ConfigEx=Day     # echoes day/night, does NOT select the profile on this camera
```

**What happens:** since #582, `_profile_mode` is derived from `ConfigEx` whenever it's present ("Day" → `"0"`). All lighting reads/writes then target `Lighting_V2[0][0][0]` (the Day profile), while the camera renders from profile **2** — so commands are accepted, the entity state reads back "on" from the inactive profile, and the physical light never fires.

**Verification:** writing `Lighting_V2[0][2][0].Mode=Manual` directly via the CGI API lights the lamp instantly (night snapshot mean brightness 54.6 → 93.8), while the integration's writes to profile 0 do nothing. On 0.9.83 (pre-#582, `Config[0]`-based) the same camera worked.

**Suggested fix** (running locally, restores correct behaviour end-to-end on this camera):

```python
config_ex = mode_data.get("table.VideoInMode[0].ConfigEx")
config0 = mode_data.get("table.VideoInMode[0].Config[0]", "")
if config0 == "2":
    # "General" profile management: profile 2 is active in all conditions and
    # ConfigEx merely echoes day/night without selecting the profile.
    self._supports_config_ex = False
    self._profile_mode = "2"
elif config_ex is not None:
    self._supports_config_ex = True
    self._profile_mode = "1" if str(config_ex).lower() == "night" else "0"
else:
    self._supports_config_ex = False
    self._profile_mode = config0 or "0"
```

i.e. only prefer `ConfigEx` when `Config[0]` doesn't indicate General mode — which should keep the #582 dual-illuminator behaviour intact (those cameras report a static `Config` list of `[0,1]`, not `2`).

Happy to test a fix build against this camera.

zios0007:
Fixes #605, and restores the read half of #582 without reintroducing it.

The day/night profile selects which `Lighting[channel][profile]` the light is read from and written to. Get it wrong and the command is accepted, the entity reads back "on" from the inactive profile, and the lamp never fires.

Two camera families disagree about which field is authoritative, so this has now been broken in both directions:

| | what the camera reports | what went wrong |
|---|---|---|
| **#582** IL series dual smart light | `ConfigEx` selects the profile; `Config[0]` stays a static `0` | reading `Config[0]` left the illuminator permanently tracking day |
| **#605** General profile management | `Config[0]=2`, one profile for all conditions; `ConfigEx` echoes day/night but selects nothing | preferring `ConfigEx` sent every write to profile 0 while the camera rendered from 2 |

#582 fixed the first by preferring `ConfigEx`. #605 was fixed by reverting it in 0.9.86 — which put #582's cameras back where they started. Neither field is right on its own.

### The change

Order them rather than choose between them:

1. `Config[0] == "2"` → profile **2**. General mode, and `ConfigEx` is noise here.
2. Otherwise `ConfigEx`, **if it names a profile we recognise** → day or night.
3. Otherwise `Config[0]`.

An unrecognised `ConfigEx` now defers to `Config[0]` instead of being read as day. Overriding a `Config[0]` that is probably right, for a string we can't interpret, is the worse of the two guesses — and it's a case my own first version got wrong.

### The per-channel read is kept, deliberately

`VideoInMode` is host-wide and returns a row per channel. The fix as suggested in #605 reads `VideoInMode[0]` directly, which would give every channel of an NVR channel one's profile — the bug fixed in #619.

That matters here because **`ConfigEx` is per-channel too**. Measured on a DHI-NVR5464-16P-EI:

```
Config[0] per channel:  {0:'0', 1:'0', ... 9:'2', 11:'0'}   <- one channel in General mode
ConfigEx  per channel:  {1:'Day', 11:'Day'}                  <- only two channels report it at all
```

Both cases on one recorder, and reading row 0 would have misread nine of the ten channels.

### Not included

`async_set_video_profile_mode` still writes `Config[0]` only, so the *other* half of #582 — forcing a profile on an IL camera so it holds — remains reverted. That's a user-invoked service rather than the polling path, and I have neither camera family to test a write against. Worth its own change by someone who does.

Note also that on a General-mode camera the service will write `Config[0]=0` or `1`, taking the camera out of General mode. That is pre-existing rather than new, but it is now visible.

### Which reintroduced bug fails which test

| mutation | tests that fail |
|---|---|
| `Config[0]` only (what main does today) | IL series takes the profile from `ConfigEx`; `ConfigEx` is read from this channel's row |
| prefer `ConfigEx` always (the 0.9.84 regression) | General profile management uses profile 2; General mode is detected from its own row |
| read `VideoInMode[0]` (as suggested in the issue) | all four channel-row tests, including the measured-NVR one |
| unrecognised `ConfigEx` means day | an unrecognised `ConfigEx` defers to `Config[0]` |

@mattmcd1 you offered to test a build against your camera — that would be worth doing before this merges, since your General-mode camera is the one I can't reproduce. @Trexano99 likewise for the IL series, if you're still around.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -1398,18 +1398,47 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
     def read_profile_mode(self, mode_data: dict) -> str:
         """Picks this channel's day/night profile out of the VideoInMode table.
 
-        The read is host-wide -- getConfig&name=VideoInMode returns a row per
-        channel -- so an NVR channel has to take its own row. Reading row 0 for
-        every channel gave the whole device channel 1's day/night profile, and
-        that profile is then what selects which Lighting[channel][profile] the
-        IR light is read from and written to.
+        The profile chooses which Lighting[channel][profile] the light is read
+        from and written to, so getting it wrong means commands are accepted and
+        nothing lights up.
 
-        Falls back to row 0, which is all a single-channel camera returns.
+        Three device behaviours have to coexist here, and two of them disagree
+        about which field is authoritative:
+
+        - **General profile management** (`Config[0]` = 2). One profile covers
+          all conditions and it is profile 2. `ConfigEx` is still present and
+          still echoes day/night, but it selects nothing -- preferring it sent
+          every write to the day profile while the camera rendered from 2 (#605).
+        - **IL series dual smart light.** The profile is chosen by the `ConfigEx`
+          string; `Config[0]` stays a static 0 whichever profile is live, so
+          reading it left the illuminator permanently tracking day (#582).
+        - **Everything else.** `Config[0]` is the profile.
+
+        The read is host-wide -- getConfig&name=VideoInMode returns a row per
+        channel -- so an NVR channel has to take its own row, falling back to
+        row 0, which is all a single-channel camera returns.
         """
-        mode = mode_data.get("table.VideoInMode[{0}].Config[0]".format(self._channel))
-        if mode is None:
-            mode = mode_data.get("table.VideoInMode[0].Config[0]", "0")
-        return mode or "0"
+        def field(name):
+            value = mode_data.get("table.VideoInMode[{0}].{1}".format(self._channel, name))
+            if value is None:
+                value = mode_data.get("table.VideoInMode[0].{0}".format(name))
+            return value
+
+        config = field("Config[0]")
+        config_ex = field("ConfigEx")
+
+        if config == "2":
+            return "2"
+        if config_ex is not None:
+            # Only act on a value we recognise. Treating anything else as day
+            # would override a Config[0] that is very likely right, for a
+            # string we do not understand.
+            named = str(config_ex).strip().lower()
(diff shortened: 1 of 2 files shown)


Trexano99:
Hi @rroller, i am sry for the introduced regression.
I hope i will be able to test it in the next days. 
I will update you once done.
Thx in advance! 😊

Issue #636, opened 2026-09-08
Move polling to RPC2 so a poll costs one login, not one per call


zios0007:
@rroller — a direction question before I write anything, because it's a change to how every read leaves the integration and I don't want to build it if you'd rather not own it.

## The problem the recent work does not solve

#577 and #603 are both, underneath, the same complaint: the integration writes a lot of entries to the device's own log. #616, #628, #630 and #632 all reduce **how many calls** happen. None of them touch **what a call costs**, and that's where the remaining complaint lives.

Every CGI call is a separate HTTP request with its own digest handshake. Measured on a DHI-NVR5464-16P-EI, the device answers `Connection: close` and reissues its nonce every time, so a cached challenge is always refused: each logical call costs two TCP connections, two HTTP requests, and a login the device records.

The user in #577 has now tested 0.9.87 and 0.9.92 on a clean VM and is switching to the ONVIF integration, because ONVIF authenticates once per session and doesn't produce the entries. That comparison is fair, and with CGI there is no version of this where we win it — the floor is one login per call.

## RPC2 removes the floor, and we already ship a client for it

`rpc2.py` has `login`, `logout`, `request` with session handling, `get_config` and `get_coaxial_control_io_status`. It's currently used only for a couple of PTZ functions on the SDT4E425.

I pointed it at my NVR today, through the integration's own client:

```
login                          2 HTTP requests
7 config reads                 7 requests, one each, no re-authentication
CoaxialControlIO.getStatus     works
logout                         1
                              --
total                         11

the same 7 reads over CGI     14 requests, each with a refused login
```

All seven succeeded — `MotionDetect`, `DisableLinkage`, `DisableEventNotify`, `SmartMotionDetect`, `Lighting_V2`, `VideoInMode`, `Lighting[0][2]` — so RPC2 can serve the whole polling surface, not just PTZ.

**The seven reads cost no additional logins.** If the session persists across polls, steady state becomes roughly one login per session instead of one per call. That is the difference between this integration and ONVIF, and it's the only lever left that changes the order of magnitude.

## What it would take

Roughly 40% of the work is not the reads:

- **Session lifecycle** — keepalive, expiry detection, re-login without every entry stampeding at once.
- **A CGI fallback** for devices without RPC2, which means two transports to maintain rather than one.
[...]

PR #639, opened 2026-09-09
Read config over one RPC2 session per host, instead of a login per call

zios0007:
Draft, as offered in #636. Opt-in per entry, default off.

## Why

Every CGI call is a separate HTTP request with its own digest handshake, so each one costs a refused login and then an accepted one. #616, #628 and #632 reduced **how many calls** happen. None of them touch **what a call costs**, and that floor — one login per call — is the whole of the complaint in #577, where the reporter has moved to ONVIF precisely because it authenticates once per session.

**Whose log this quietens is device-class specific, and that is now measured rather than assumed.** The request cost is the same everywhere — both my NVR and my direct VTO doorbell answer `[401, 200]` on every call with a cached challenge refused. What differs is what the device *writes down*. Recorders log a line per call; the direct SD49225XA-HNR tested at length in #577 does not — integration setup produced a single entry, and only writes and their follow-up reads ever appeared. So this change is worth having on a recorder, and on a direct camera it is a request-count change you will not see in the log. I had earlier stated the per-call logging as a general property of Dahua devices; that was wrong, and the correction is the reporter's measurement, not mine.

RPC2 authenticates once per session too, and this repo already ships a client for it.

## What it does

Routes `getConfig` reads over **one login per host**, held open by a keepalive, and gives it back when the last entry unloads.

```
RPC2    7 HTTP requests for 7 config reads on an open session
CGI    14 requests for the same 7, each carrying a refused login
```

Verified on a DHI-NVR5464-16P-EI: three clients for one host produce **one login** and a refcount of three; reads are byte-identical to CGI across five configs; two closes leave the session up; the third cancels the keepalive and closes it; a fourth is a no-op.

## The translation is exact

Every accessor here reads CGI's flat shape — `table.Lighting_V2[0][1][0].Mode` — while RPC2 answers in nested JSON. Checked across seven configs and **4,647 keys**: same keys, same values, zero differences.

**Nulls are dropped, and that is not cosmetic.** CGI omits an absent entry; RPC2 sends JSON `null`. Keeping them would give every channel a `SmartMotionDetect` row — the phantom switch removed in #635, straight back.

## Design notes

**One session per host, keyed `(address, username)`.** A session id is obtained with the password and scoped to that user's rights, so it follows the digest state's rule rather than the connector's. Eleven channels of one NVR cost one login, which is most of the point.

**No lock.** The login future is registered before it is awaited, so concurrent first reads share the one in flight. That is `_SharedRead`'s move; this integration has no `asyncio.Lock` anywhere and ordering is its house style.

**Keepalive at the device's own interval**, from `keepAliveInterval` in the login reply, less five seconds — the same shape as the existing VTO keepalive. Started on first use, never in the constructor: a refcount survives a failed setup, a task started in a synchronous constructor does not, which is the leak #620 was about.

**Release is refcounted and idempotent**, guarded by its own flag rather than by whether a socket closed cleanly — `async_stop` is reachable three ways and a second release would close the session under the other entries. At zero the keepalive is cancelled *and awaited* before the logout, so it cannot fire one more request at a session being closed.

**The "RPC2 does not work here" verdict is per host.** Eleven channels each rediscovering it is eleven failed logins against a device that has just said it cannot.

## Two bugs this fixes in the first draft

- `close()` cleared the aiohttp session but not the RPC2 client, so afterwards it handed out a client bound to a closed socket.
- `close()` never logged out. A Dahua box has a finite session table, and the per-call PTZ helpers already log out.

Also dropped `enable_cleanup_closed` from the RPC2 session — aiohttp ignores it on every Python Home Assistant runs and warns once per connector, which is why the shared connector dropped it in #620. The unsafe cookie jar stays; RPC2 sets a cookie against a bare IP.

## The keepalive is load bearing, measured

Two runs disagreed at first, so I ran the comparison properly. 400 seconds, same device, same credentials:

```
idle, no keepalive       -> session dead
keepalive every 50s      -> session alive
```

So the timeout is idle based, somewhere between 200 and 400 seconds — an earlier 200s idle survived, 400s did not — and the keepalive holds it open indefinitely.

It matters in ordinary operation, not only at long poll intervals. Since #628 a config read is cached for 300 seconds, so a host can go a full 300 seconds making no RPC2 request at all because every read is a cache hit. That sits inside the expiry window. Without the keepalive the session would lapse during normal quiet polling and every cache expiry would pay for a fresh login — which is the cost this change exists to remove.

It costs one request per minute per host, against seven per poll removed.

## Not here, deliberately

- **Writes.** Reads are the bulk and the safe half.
- **Flipping the default.** Everything measured on this integration lately varies by model, and I have one NVR and one VTO.

Diagnostics report the session as present or absent with a refcount, never the id, under the same rule as the digest state.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -45,6 +45,7 @@ from .const import (
     CONF_AUTO_DETECT_CHANNEL,
     CONF_USE_HTTPS,
     CONF_SCAN_INTERVAL,
+    CONF_USE_RPC2,
     CONF_NVR_ACTIVE_DETERRENCE,
     DEFAULT_SCAN_INTERVAL,
     MIN_SCAN_INTERVAL,
@@ -629,7 +630,8 @@ class DahuaDataUpdateCoordinator(DataUpdateCoordinator):
 
         # The client used to communicate with Dahua devices
         self.client: DahuaClient = DahuaClient(username, password, address, port, rtsp_port, self._session,
-                                               use_https)
+                                               use_https,
+                                               use_rpc2=entry.options.get(CONF_USE_RPC2, False))
 
         # self.config_entry = entry
         self.platforms = []
--- a/custom_components/dahua/client.py
+++ b/custom_components/dahua/client.py
@@ -2,6 +2,7 @@
 import logging
 import re
 import socket
+from contextlib import suppress
 import asyncio
 import time
 import aiohttp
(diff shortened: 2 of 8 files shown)


zios0007:
Out of draft. When you last saw this it was a sketch with a *Known gaps* section; those are now built and measured, so it is a real proposal rather than something to look at.

What changed since:

- **One session per host, not per entry.** Keyed `(address, username)` — a session id is obtained with the password and scoped to that user's rights, so it follows the digest state's rule rather than the connector's. Eleven channels of an NVR now cost one login instead of eleven, which was most of the point.
- **A keepalive**, at the interval the device states in its own login reply. This turned out to be load bearing rather than insurance: with the 300s config cache from #628, a host can go a full 300 seconds making no RPC2 request at all because every read is a cache hit, and that sits inside the expiry window. Measured over 400 seconds — idle, the session dies; keepalive every 50s, it survives.
- **Refcounted, idempotent release** with logout, and the keepalive cancelled *and awaited* before the session closes so it cannot fire one more request at it. `async_stop` is reachable three ways, so a second release must not close the session under the other entries.
- **Two bugs fixed from the sketch:** `close()` cleared the aiohttp session but not the RPC2 client, so it then handed out a client bound to a closed socket; and it never logged out, on a device with a finite session table.

Verified on a DHI-NVR5464-16P-EI: three clients for one host produce one login and a refcount of three, reads are byte-identical to CGI across five configs, and the translation matches across seven configs and 4,647 keys.

Scope is unchanged and deliberate: **reads only, default off.** Writes stay on CGI, and I would not flip the default on one NVR's evidence — everything measured on this integration lately has varied by model.

No rush on this one. It is the kind of change that is better sitting for a while than merged quickly.

zios0007:
Narrowed one claim in the description, because #577 produced a measurement that contradicts it.

I had written that a CGI call costs a login the device logs, as though that were true of Dahua devices generally. The **request** cost is general — my NVR and my direct VTO doorbell both answer `[401, 200]` on every call, cached challenge refused. What is not general is the logging. @JimGermany's direct SD49225XA-HNR wrote a single entry for integration setup and nothing afterwards except writes and the reads that follow them, over repeated tests.

So the honest claim is narrower than the one I opened with: this quietens **recorders**. On a direct camera it is a request-count change you will not see in the device log. That does not change the code or the measurements, only what the PR promises.

Nothing else has changed since you last looked — CI is green and it is still opt-in, default off.

rroller (approved):

PR #641, opened 2026-09-10
Add profile sensor and firmware build date to diagnostics

roalvesrj:
- Add DahuaProfileSensor: exposes day/night lighting profile (0=Day, 1=Night, 2=Scene) with raw profile_number attribute
- Add get_build_date() to coordinator: parses build date from firmware version string (e.g. '2.800...build:2020-06-05')
- Expose build_date on DahuaFirmwareVersionSensor via extra_state_attributes
- Include build_date in diagnostics device block


zios0007 (commented):
Not a maintainer — just a contributor who has been in these files lately, so take this as a read rather than a verdict. The build date part looks right to me: the parse handles the absent case, and `extra_state_attributes` on the base returns a fresh dict each call, so adding a key to it is safe.

Three things on the profile sensor.

**1. On a device without profile support it will report "Day" forever, including at night.**

`_profile_mode` is initialised to `"0"` (`__init__.py:649`) and only ever updated inside

```python
if self._supports_profile_mode and not self.is_doorbell():
```

(`__init__.py:959`). Everywhere else it keeps the initial value, so `get_profile_mode()` returns `"0"` and the sensor renders **Day** permanently — on a doorbell, and on any camera where `_supports_profile_mode` came back false. That is worse than no sensor, because a wrong value looks like a working one.

The probe already runs at startup, so gating costs no extra request:

```python
if coordinator.supports_profile_mode():   # or the attribute directly
    sensors.append(DahuaProfileSensor(coordinator, entry))
```

or keep it unconditional and return `None` when unsupported, so it shows as unavailable rather than as Day. Either is fine; the current behaviour is the one to avoid. This is the same shape as #635 and #638 — the recent pattern in this repo has been to give an entity only to devices that actually have the thing.

**2. It has no `entity_category`.**

`DahuaFirmwareVersionSensor` and `DahuaSerialNumberSensor` are both `EntityCategory.DIAGNOSTIC` (set in #634). Without it the profile sensor lands in the device's main entity list rather than under Diagnostics, which is inconsistent within the same file.

**3. The name is missing the device prefix.**

The other two use a `name` property returning `self._coordinator.get_device_name() + " ..."`. This one sets `_attr_name = "Profile"` flat, and the integration doesn't use `_attr_has_entity_name`, so nothing prefixes it. On a multi-camera or NVR setup every entry produces an entity called "Profile" and HA disambiguates by suffixing the entity_id — `sensor.profile`, `sensor.profile_2`, and so on, in whatever order they set up. `get_device_name() + " Profile"` matches the file and keeps the ids stable.

Also worth flagging for scheduling rather than for you to fix: I have #639 open against `__init__.py` and `diagnostics.py`, which this also touches. Whichever lands second will need a rebase — mine is the one that should absorb it if it comes to that, since yours is the smaller diff.

zios0007:
Correction to my last paragraph: #639 has since merged, and I tested this branch against the new `main` — it merges clean, no conflict in either `__init__.py` or `diagnostics.py`. My rebase warning was wrong, so please ignore it; nothing here needs reordering.

The three points about the profile sensor still stand, the first one especially — on a device where `_supports_profile_mode` is false it will report Day permanently.

roalvesrj:
@zios0007, thanks for the careful review and for the heads-up about #639 — good to know it merges clean, no reordering needed here.

All three points on the profile sensor are addressed:

1. Gating on the capability — exactly the one you emphasized, and you're right that it would have reported "Day" forever. I added coordinator.supports_profile_mode() and async_setup_entry in sensor.py now only appends DahuaProfileSensor when it's True. Doorbells and cameras without selectable profiles get no profile sensor at all.
2. EntityCategory.DIAGNOSTIC — applied to DahuaProfileSensor.
3. Device-prefixed naming — switched to name/unique_id properties following get_device_name() + " Profile" / get_serial_number() + "_profile", consistent with the other sensors (no has_entity_name).

Also extended tests/dahua/test_entity_categories.py to cover the new behavior, including the "no support → no sensor" case so this doesn't regress.

zios0007:
Checked all three against the branch rather than taking them on trust — they're right.

The gating is correct on the detail that mattered: `_supports_profile_mode` is only assigned inside `if not is_doorbell:` (`__init__.py:920`), so a doorbell keeps the `False` it was initialised with and gets no sensor. Your docstring's claim holds.

The ordering also works out, which was the other way this could have gone wrong: `async_config_entry_first_refresh()` runs at `__init__.py:208`, before `async_forward_entry_setups()` at 230, so the probe has completed by the time `sensor.async_setup_entry` asks. A supported camera won't be skipped because the answer wasn't ready yet.

`name` and `unique_id` now match the other two sensors, and the category is set. Nothing further from me.

One thing outside the code: the checks aren't reporting here, which I think is because the PR head is your fork's `main` branch. @rroller may need to approve the workflow run before this can go green.

rroller (approved):

rroller:
Looks like the tests failed

roalvesrj:
Fixes applied and PR synced.

zios0007:
@roalvesrj heads up — **CI has not run on your latest push, so the fix is not actually verified yet.**

The only run this PR has ever had was on `95addb0d`, and that is the failure @rroller saw. There are **zero check runs** on `58d1273c`, and between the two there is only an upstream merge, no fix commit. So as things stand the same failure would most likely still be there.

It is a small one, and it is not in your production code.

```
AttributeError: '_Coordinator' object has no attribute 'get_build_date'
custom_components/dahua/diagnostics.py:131
```

**Why `_safe` does not catch it:** `_safe(coordinator.get_build_date)` resolves the attribute *before* `_safe` is ever called, so a missing method raises at lookup time rather than inside the guard. Every other line in `_device_block` has the same shape and gets away with it only because the test double defines those methods.

**The fix is one method on the stub** at `tests/dahua/test_diagnostics.py:55`. It already has `get_profile_mode`, which is why the profile sensor half of your change passes — `get_build_date` is the one that was missed:

```python
    def get_build_date(self):
        return "2020-06-05"
```

**On why CI keeps not running:** this PR's head is your fork's `main` branch, and workflow runs for it need @rroller to approve them each time. That is why the last two pushes show no checks at all rather than a red X — worth asking him to approve the run once you have pushed the stub fix, otherwise it will look stalled again.

The three review points from before still hold up on my read of the branch; this is purely the test double.

roalvesrj:
Good catch, thanks. Added *get_build_date()* to the test double and pushed. Checks are pending your approval.

Issue #644, opened 2026-09-11
Event stream closed by device after recent firmware update on IMX-C cameras


physisheckman:
Hello,

​After a recent firmware update on my Intelbras IMX-C cameras, the integration stops receiving motion and human detection events. The event stream keeps disconnecting, and the logs show that the connection is being actively closed by the devices.

​Here are the details from the Home Assistant logs:
Logger: custom_components.dahua
Source: custom_components/dahua/__init__.py:552
Integration: Dahua (documentation, issues)

Event stream for 192.168.100.103 ended unexpectedly: Event stream to 192.168.100.103 closed by the device
Event stream for 192.168.100.108 ended unexpectedly: Event stream to 192.168.100.108 closed by the device
Event stream for 192.168.100.116 ended unexpectedly: Event stream to 192.168.100.116 closed by the device
Event stream for 192.168.100.115 ended unexpectedly: Event stream to 192.168.100.115 closed by the device
Event stream for 192.168.100.118 ended unexpectedly: Event stream to 192.168.100.118 closed by the devicea

Additional Context:
​Integration Version: 0.9.93
​Home Assistant Version: 
Método de instalação
Home Assistant OS
Core
2026.9.1
Supervisor
2026.09.0
Operating System
18.2
Front-end
20260826.6
​Camera Line: Dahua / IMX-C series

​It looks like the new firmware version modified how HTTP/TCP persistent connections or keep-alives are handled, causing the device to drop the event stream socket. Is there a way to handle automatic reconnections or adjust the keep-alive parameters for this new firmware behavior?

​Thank you for your time and for maintaining this great integration!

zios0007:
@physisheckman thanks — the log lines and the firmware detail are the useful half of this already.

**Reconnection is already automatic**, so that part isn't missing. The stream runs in a loop that re-attaches after every ending, and since 0.9.93 that warning is deliberately logged **once per outage** rather than once per retry. Five lines for five cameras means each one had an episode, not that each one is failing repeatedly — the repeats are at debug level.

**What decides whether you're actually losing events is how long each stream lives before it's closed**, because that sets the retry delay:

| Stream lived | Next attempt |
|---|---|
| 60s or more | immediately |
| 10–60s | ~10s |
| under 10s | 60s, doubling per consecutive instant death, capped at 10 min |

[...]

PR #645, opened 2026-09-11
Decide the event stream retry on whether the device spoke, not how long it lasted

zios0007:
Two halves of one idea: **let the retry decision look at whether the device actually said anything**, instead of at how long the socket lasted. Duration cannot tell a working stream from a broken one in either direction.

## Half one: a device that is talking should not be backed off

`event_stream_retry_delay` decided on duration alone, so a stream that lasted under ten seconds was treated as a device dying on contact and backed off — 60s, doubling, capped at ten minutes.

That is right for a device refusing to attach. It is wrong for firmware that hangs up early, and the two are indistinguishable by duration while meaning opposite things. Against a device that delivers events and is closed at eight seconds:

```
close #   retry delay (s)
1         54
2         117
3         227
4         453
...
10        602

after 10 closes: ~74 min elapsed, now asking every ~10 min
```

A camera detecting motion perfectly well ends up unwatched for most of the hour. Treated as short-but-alive it retries every ~11s instead.

## Half two: a device that says nothing is broken, however long it holds the socket

This one is worse, and #644 is what turned it up.

The stream asks for a heartbeat every five seconds and gives up after sixty seconds of silence. That read timeout raises `aiohttp.ServerTimeoutError` — which **is** a `TimeoutError`, the same exception the deliberate hourly recycle raises through `wait_for`:

```python
>>> issubclass(aiohttp.ServerTimeoutError, asyncio.TimeoutError)
True
```

Both landed in the branch meaning *this stream was working*. So a subscription that attached and delivered nothing was recorded as healthy — no warning, failure counters cleared, reconnected at once. The device produces no events at all, indefinitely, while the log says nothing is wrong and the integration pays a login a minute for it.

In #644 a firmware update made cameras stop matching any event when too many codes were subscribed at once. The endpoint still accepted the subscription and held it open. It took the reporter several rounds to find, because nothing anywhere said the stream was mute. There is now a warning that names the likely cause.

**Silence is never innocent here.** `_restart_if_needed` only starts a stream when some event is wanted, so there is no correctly-subscribed device that sends nothing — a camera with nothing to report still heartbeats.

## What is unchanged

- A stream that delivered and lived a healthy while still reconnects immediately.
- A stream that said nothing and died instantly still backs off exactly as before.
- The new argument defaults to `False`, so every existing caller keeps the old behaviour.
- A heartbeat counts as the device talking, deliberately: the flag is set before the parse that discards heartbeats.

## Verification

The delay function is pure, so the tests call it directly, and I checked them against the previous behaviour rather than only against the new one — the silent-stream tests fail on the code as it stood before this change, while the two that pin healthy behaviour pass on both. That split is the point: if the preservation tests had also flipped, something had broken.

One assertion from the first half of this PR is reversed here. It claimed a stream living an hour reconnects at once regardless of whether it delivered anything. That is precisely the bug.

### DIFF
--- a/custom_components/dahua/__init__.py
+++ b/custom_components/dahua/__init__.py
@@ -95,9 +95,18 @@ PROBE_FAILED = (ClientError, TimeoutError)
 PROBE_REFUSED = (ClientResponseError, TimeoutError)
 
 
-def event_stream_retry_delay(lived_seconds: float, consecutive_failures: int = 0) -> float:
-    """How long to wait before re-attaching, given how long the stream lasted."""
-    if lived_seconds < 10:
+def event_stream_retry_delay(lived_seconds: float, consecutive_failures: int = 0,
+                            received_data: bool = False) -> float:
+    """How long to wait before re-attaching, given how long the stream lasted.
+
+    `received_data` is whether the device sent anything at all on this attach --
+    an event or a heartbeat. It separates the two things a short stream can mean.
+    A device that refuses attach and one whose firmware hangs up after eight
+    seconds of perfectly good events look identical by duration, and only the
+    first should be backed off: the second is working, and backing it off to ten
+    minutes is how a camera that still detects motion stops reporting any.
+    """
+    if lived_seconds < 10 and not received_data:
         # Double per successive instant death, so a device that is refusing
         # attach gets asked less often the longer it keeps refusing.
         doublings = max(0, consecutive_failures - 1)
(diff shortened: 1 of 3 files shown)


zios0007:
Expanded this while #644 was being diagnosed, and retitled it accordingly — it was "don't back off a device that is talking", it is now also "don't call a silent stream healthy". Same one-line idea underneath, so it seemed better as one change than two.

The addition is the more serious of the two. `sock_read` fires after sixty seconds of silence and raises `ServerTimeoutError`, which is a `TimeoutError` — the same exception the deliberate hourly recycle raises. Both were landing in the branch that means the stream was working, so a subscription delivering nothing was recorded as healthy and reconnected forever, silently.

@physisheckman's #644 is exactly that: firmware stopped matching events when too many codes were subscribed, the endpoint kept accepting the subscription, and nothing in the log said the stream was mute. That thread took several rounds to reach a cause the integration could have named. It now warns once and points at the event-type count.

CI is green. Happy to split it back into two PRs if you would rather review them separately.

rroller (approved):

PR #653, opened 2026-09-13
Use the aiodahua library instead of a vendored client

brianegge:
Replaces the vendored client with a dependency on [aiodahua](https://github.com/brianegge/aiodahua), leaving a ~60 line shim that preserves the original constructor signature.

**−1588 / +132 lines.** `__init__.py`, `config_flow.py`, `light.py` and `switch.py` are untouched — they import exactly what they did before.

## Why

Two independent async Dahua clients existed: this one, and the one inside `dahua-mcp`. Their protocol knowledge was *complementary rather than duplicated* — this client knew events, RTSP and speaker audio; the other knew config-write quirks, storage and recording search. Neither was a superset. Extracting a shared library means that knowledge is written down once, versioned and tested on its own.

## This is not a behavioural rewrite

Every method kept its name, signature and return shape, including the md5 serial fallback and the device-literal `table.` keys. `aiodahua` carries all 64 methods across verbatim; the diff here is deletion, not translation.

Three subtleties I made sure to preserve:

- **`parse_kv` does not strip the `table.` prefix.** The library originally did. Callers index the literal response, so renaming keys would have broken them silently. Stripping is now opt-in.
- **Exception types inherit from the aiohttp ones they replace.** This matters more than it looks: capability detection here probes an endpoint and reads `ClientResponseError` as "unsupported", and the reauth path reads `exception.status`. So `DahuaResponseError` *is* a `ClientResponseError` carrying `status`, `DahuaNotSupportedError` is a 400, `DahuaAuthError` is a 401, and `DahuaTimeoutError` is still a `TimeoutError`. Every existing `except` fires as before.
- **`ClientResponseError` is not flattened into a connection error.** My first attempt did, which would have stopped `async_get_system_info`'s md5 fallback from ever firing. Caught by the existing tests.

## Bug fixed along the way

The three services that write free user text — `set_channel_title`, `set_text_overlay`, `set_custom_overlay` — interpolated straight into the query string:

| Input | Before | After |
|---|---|---|
| `Lot #2` | truncated to `Lot ` at the `#` | correct |
| `Bill & Ted` | HTTP 400 | raises `DahuaValueError` (firmware cannot represent `&` at all) |
| `""` (clear) | silently did nothing, **and still passed the `OK` check** | genuinely clears |

The `\|` field separator now travels as `%7C`; the device decodes it back. Confirmed on a live NV4116-HS, along with every other change here.

## Tests

**471 pass.** Transport tests now patch `aiodahua.client.DigestAuth`, since that is where the transport lives, and the overlay assertions expect the encoded URL.

`test_coordinator_init::test_setup_success` errors on a lingering-thread assertion inside `pytest_homeassistant_custom_component` — I verified it errors identically on the branch point, so it is not from this change.

## Note

`manifest.json` now declares `aiodahua==0.2.0`. It is on GitHub but **not yet published to PyPI**, so this cannot merge until it is.


brianegge:
Opened against the wrong repository by mistake — this was meant for my own fork. Closing straight away; apologies for the noise.

PR #654, opened 2026-09-13
Say so when the lighting scheme will not let the light come on

zios0007:
Follow-up to #652, from @roby-esc's testing of 0.9.95 on #647. They confirmed #652 works — the white light now physically responds — and in the same report found that it is not sufficient on their camera.

## The remaining problem

Smart Dual Light cameras decide separately, in `LightingScheme`, which emitter they are willing to use:

```
LightingScheme[0][0].LightingMode=AIMode      -> illuminator writes correctly, light stays off
LightingScheme[0][0].LightingMode=WhiteMode   -> same command, light physically comes on
```

Nothing about the integration changed between those two lines. They switched the mode in the camera's own web interface.

So a write can land on exactly the right `Lighting_V2` row, store correctly, be read back correctly, and still do nothing visible. From inside Home Assistant that is unfalsifiable: the entity reports on, the light is off, and there is no way to tell that from a working setup. It cost them an evening and a great deal of careful testing to establish.

## What this does, and what it deliberately does not

It warns. It does not change the scheme.

Setting `LightingScheme` to `WhiteMode` would make the command work, and would also turn Smart Dual Light off on someone's camera because they tapped a light in Home Assistant. @roby-esc raised the same concern unprompted — *"the previous scheme should be preserved or Smart Dual Light could be disabled unexpectedly"* — and I agree. The integration says what happened; the user decides.

## On the extra request

`LightingScheme` is read when the illuminator is switched on, not on every poll. Light commands are rare and user initiated, whereas polls are what #616, #628, #632 and #639 have been spent making cheaper, and a warning most devices never need does not justify a request every cycle. The read is wrapped: a camera that has no such table fails the lookup, is ignored, and stays silent — which is every camera older than this.

Only `WhiteMode` is known to pass the white light through, so anything else is treated as blocking, including modes that do not exist yet. The alternative is silence on the exact firmware this is for.

## Verification

The rule is pure, so the tests call it directly, and I checked them against the previous behaviour rather than only the new one: substituting "nothing ever blocks" fails the three cases that report a block and passes the four that pin the silence.

`test_light.py`'s fake client gained the method, deliberately — without it the `except` branch would swallow an `AttributeError` and the tests would pass while exercising nothing. There is a test that a camera reporting no scheme still switches on cleanly and that the scheme is actually consulted.

## Still open on #647

Two further findings from that report are not addressed here: the Infrared entity writes the legacy `Lighting` table while that camera physically uses the `Lighting_V2` `InfraredLight` row, and `VideoInMode[0].ConfigEx` reports `Day` on that firmware while the camera is using the Night profile. The second one limits #638 and I would want more than one camera before touching that ordering again.


rroller (approved):

rroller:
Feel free to merge when you are ready

zios0007:
Thanks — rebased onto current `main` and green, so this is ready whenever you want it. Please merge at your convenience; squash to match the rest of the history.

It had gone conflicted against #659 and #660 after they landed. Both conflicts were additive rather than contradictory, so both sides are kept:

- `__init__.py` — `illuminator_brightness_bank` and `scheme_blocking_white_light` are separate module-level helpers that happened to land at the same anchor.
- `light.py` — one side passes the resolved brightness bank, the other adds the scheme check after the write. The turn-on path now does both.

Worth noting for #656, since it touches the same area: that PR writes the scheme with `configManager.setConfig`, so the stored value changes and `getConfig` reflects it. This warning reads the stored `LightingMode`, so the two agree. They would **not** have agreed if it used `setTemporaryConfig` — @roby-esc measured on #647 that a temporary override is invisible to both CGI and RPC2 reads, which would have made this warn while the light was physically on. If #656's transport ever changes, this is the thing to re-check.

zios0007:
I would have, but I do not have merge rights on the repo — GitHub returns:

```
zios0007 does not have the correct permissions to execute `MergePullRequest`
```

So it needs you, or write access, whichever you prefer. No urgency either way: it is rebased, green, and the only thing it changes is that a user gets told why an illuminator command that looked successful produced no light.

PR #655, opened 2026-09-14
Add IPC-Color4M-TZ security light support

kreucher:
## Summary

Expose the existing Dahua Security Light entity for the `IPC-Color4M-TZ` camera family.

Two physical cameras were tested. On both, the integration's existing coaxial security-light API controls the alternating red/blue active-deterrence LEDs, despite the API naming this output `WhiteLight`.

## Problem

The cameras support the same `coaxialControlIO.cgi` active-deterrence control already implemented by the integration, but `supports_security_light()` does not recognize this model family. Home Assistant therefore does not create a Security Light entity.

The API's `status.WhiteLight` field is misleading on this model: it reports the separate red/blue warning output, not the camera's steady white illuminator.

## Changes

- Add the `IPC-Color4M-TZ` model prefix to the existing security-light capability gate.
- Normalize model matching to uppercase while preserving the existing supported models.
- Add focused tests for the new model family, case-insensitive matching, and an unrelated unsupported model.

No new entity type or API call is introduced.

## Validation

- Physically verified alternating red/blue output on two `IPC-Color4M-TZ` cameras.
- Direct API verification on both cameras: `coaxialControlIO.cgi`, type `1`, IO `1` enabled the red/blue output, IO `2` disabled it, and `status.WhiteLight` reported the resulting state.
- Manually toggled both new Home Assistant Security Light entities and physically verified correct on/off behavior. Each red/blue output worked independently and while its separate steady white Illuminator was also on.
- Exercised the existing armed/night North and South automation action paths against the deployed entities. Each automation activated only its matching Security Light, and the shared failsafe turned both off.
- Tested firmware on both cameras: `3.140.0000000.21.R`, build `2025-12-06`. This is newer than the `2025-10-22` firmware currently listed for this model on EmpireTech's public download page but came pre-installed with each camera.
- Deployed combined fork head `aa4eace` as a Home Assistant canary based on integration `0.9.95`; both cameras expose available, independent Illuminator and Security Light entities.
- `pytest -q`: 571 passed on the isolated security-light PR head using Python 3.13.
- `git diff --check upstream/main...HEAD`: clean.

## Development disclosure

AI assistance was used during the investigation, implementation, test automation, evidence analysis, and drafting of this report. All reported measurements came from the stated test systems. Nicholas Kreucher (`@kreucher`) performed the physical camera validation and reviewed the observed behavior, validation evidence, proposed change, and final report before submission.


zios0007 (commented):
Not a maintainer — a read, not a verdict. This one looks right to me and I would take it as-is.

It follows what `supports_security_light()` already is: a list of model strings, alongside `-AS-PV`, `AD410`, `DB61I` and `IP8M-2796E`. Adding one more in the same shape, verified on two physical cameras, with tests including the suffixed-model case that OEM firmware actually produces. Nothing clever, which is the right instinct here.

The observation in your comment is worth more than the diff, so I would keep it there where someone will find it:

> `Type=1` drives the alternating red/blue active-deterrence LEDs, despite the CGI status field calling the output `WhiteLight`.

That matches what I measured on a DHI-NVR5464-16P-EI: the `WhiteLight` name in the coaxial API is not a reliable statement about what the hardware is. I also found the `coaxialControlIO` 200/400 split does **not** discriminate capability on that recorder — seven channels answer 200 and only two have a white light at all — so probing is not available as an alternative here and a model list really is the honest option for this entity.

One adjacent note rather than a request: the `Lighting_V2` illuminator path learned recently that light **indices** vary by model, and #652 now resolves that from the device's own `LightType` instead of assuming index 0. Different entity and different API, so nothing to change here — just worth knowing the two lighting paths have started diverging in how much they trust the device's own naming.

rroller (approved):

PR #656, opened 2026-09-14
Control IPC-Color4M-TZ white illuminator

kreucher:
## Summary

Correct the existing Illuminator entity for the `IPC-Color4M-TZ` camera family by coordinating the camera's `LightingScheme` and `Lighting_V2` configuration tables.

## Problem

The existing Illuminator implementation writes only `Lighting_V2`. These cameras accept and read back that change but produce no physical white light.

Physical isolation testing established that the steady white illuminator requires both `LightingScheme=WhiteMode` and the active profile's declared white-emitter row in `Lighting_V2` set to `Manual`. Applying either setting alone produced no light.

This output is separate from the red/blue active-deterrence LEDs exposed by the camera's coaxial-control API as `status.WhiteLight`.

## Changes

- Probe `LightingScheme` before exposing this model-specific Illuminator backend.
- Commit the complete `LightingScheme` and `Lighting_V2` tables together over RPC2 when controlling the Illuminator.
- Select `WhiteMode` and configure the active profile's declared white emitter when turning on.
- Restore the previous lighting scheme when turning off without overwriting emitter preferences.
- Report the Illuminator as on only when both the scheme and white-emitter mode indicate that the steady white light is active.
- Add focused tests for table mutation, RPC2 payloads and failures, entity routing, capability gating, state, and brightness.

Other camera models retain the existing Illuminator behavior.

## Validation

- Physically isolated the required configuration on an `IPC-Color4M-TZ`: `LightingScheme=WhiteMode` alone and `Lighting_V2` `Manual/100` alone did not illuminate the lamp; applying both together did.
- Manually toggled the Home Assistant Illuminator entities on two physical cameras and verified correct steady-white on/off behavior.
- Each Illuminator worked independently and concurrently with its separate alternating red/blue Security Light.
- Tested firmware on both cameras: `3.140.0000000.21.R`, build `2025-12-06`. This is newer than the `2025-10-22` firmware currently listed for this model on EmpireTech's public download page but came pre-installed with each camera.
- Deployed combined fork head `aa4eace` as a Home Assistant canary based on integration `0.9.95`; both cameras expose available, independent Illuminator and Security Light entities.
- `pytest -q`: 647 passed on rebased head `cb85d64` using Python 3.13.13.
- `git diff --check upstream/main...HEAD`: clean.

## Development disclosure

AI assistance was used during the investigation, implementation, test automation, evidence analysis, and drafting of this report. All reported measurements came from the stated test systems. Nicholas Kreucher (`@kreucher`) performed the physical camera validation and reviewed the observed behavior, validation evidence, proposed change, and final report before submission.


zios0007 (commented):
Not a maintainer, just someone who has been in these files this week — take this as a read rather than a verdict. The complete-table approach is right and the validation is careful: `deepcopy`, checking `LightType` before touching a row, raising on malformed tables rather than writing something half-formed. Partial writes to these tables are a known trap.

One thing I think is a real bug, and two worth weighing.

## 1. `AIMode` on off is wrong for a camera that did not start there

```python
scheme_row["LightingMode"] = "WhiteMode" if enabled else "AIMode"
```

This does not restore the previous scheme, it imposes one. A camera sitting in `InfraredMode` gets moved to `AIMode` the first time someone turns the Home Assistant light off, and never gets moved back.

That is not hypothetical. Over on #647, @roby-esc tested exactly this on a `DH-IPC-HFW3449E-S-IL`: their Night profile was `InfraredMode`, and **capturing the previous value and restoring it worked from both starting modes** — from `InfraredMode` without wrongly landing in `AIMode`, and from `AIMode` with the full Smart Dual Light sequence intact afterwards, human detection and all.

You are already reading the complete `LightingScheme` table immediately before writing it, so the previous `LightingMode` is in your hand. Saving it and putting it back on `enabled=False` looks like a small change to `lighting_scheme_illuminator_tables`.

## 2. This runs over RPC2, which is off by default

`async_set_lighting_scheme_illuminator` goes through `_shared_rpc2()`. The RPC2 transport landed in #639 as an **opt-in option, default off**, so on a normal install there is no session — and this would either open one behind the user's back or fail, depending on how `_shared_rpc2` behaves for them. Either way it is worth being explicit about, because a light that works only when an unrelated checkbox is ticked is hard to support.

Your docstring says partial CGI writes are accepted but do not light the emitter. Worth separating two things there: is the requirement **CGI vs RPC2**, or is it **both tables in one transaction**? @roby-esc found plain CGI `setConfig` changed `LightingScheme` and physically switched their light — so the transport may not be the active ingredient, and atomicity may be. If CGI can carry both tables, this reaches every install instead of the few with RPC2 enabled.

## 3. The model gate will not age well

`m.startswith("IPC-COLOR4M-TZ")` scopes the feature to your cameras, which is the safe call for a first cut. But the same need is now reported on `DH-IPC-HFW3449E-S-IL` (#647) and `IPC-HDW3849HP-AS-PV` — and all three are firmware **3.140**, as is #546. That looks like a firmware generation rather than a model family.

You already do the model-independent thing one level down by checking `LightType == "WhiteLight"`. Gating on the *presence of a usable `LightingScheme` row* would pick up the other 3.140 cameras for free. Not a blocker for this PR, but the model list will keep growing otherwise.

## Housekeeping

#654 is open and touches `light.py` and `tests/dahua/test_light.py` as well — it warns when `LightingScheme` is not `WhiteMode`, which is the read-only half of what you are doing here. Whichever lands second will want a rebase, and if this one lands the warning should probably be scoped to devices this path does not cover.

kreucher:
Thanks for the careful review. You’re right that forcing `AIMode` on turn-off is incorrect for a camera that began in `InfraredMode`. I’ll revise this to preserve the pre-takeover mode per channel/profile, avoid replacing it during repeated brightness changes, and restore it on a successful turn-off. If no saved mode exists—for example after an HA restart—the off path will not guess a scheme; it will leave the current scheme unchanged and set the white emitter to `Off`.

`_shared_rpc2()` does lazily create a session regardless of the RPC2 polling option, so I’ll make that distinction explicit and use CGI for capability/state reads. I plan to retain RPC2 for the user-initiated two-table multicall because that is the path physically validated on these `IPC-Color4M-TZ` cameras. The CGI result on #647 establishes that `LightingScheme` can be changed independently on that model, but not that the coordinated two-table operation is equivalent on this firmware.

I’m going to keep the model gate in this PR. Presence of `LightingScheme` plus a `WhiteLight` row establishes compatible-looking data, but not equivalent physical behavior, and #647 still has separate active-profile and infrared issues. Broader capability-based support would be better as a follow-up with hardware validation.

I’ll also rebase around #654 once it lands so devices handled by this path do not receive the read-only blocking warning.

AI assistance was used to analyze the review and draft this response. Nicholas Kreucher (`@kreucher`) reviewed and approved it before posting.

rroller:
Sorry this has conflicts, do you mind fixing this

kreucher:
Fixed—rebased onto current `main`, resolved the overlap with #659/#660, and all checks are passing. PTAL when convenient.

AI assistance was used for the rebase, conflict resolution, validation, and drafting this response. Nicholas Kreucher (`@kreucher`) reviewed the result before posting.

zios0007 (commented):
The capture-and-restore is in and it reads correctly to me — `restore_mode` threaded through, `_lighting_scheme_restore_modes` keyed by channel and profile, and the mode captured before the override rather than assumed. That was the point that mattered, and @roby-esc's testing on #647 since then confirms restoring the captured value works from both `InfraredMode` and `AIMode`, with the full Smart Dual Light sequence intact afterwards.

One thing I would still raise, about this branch:

```python
elif restore_mode is not None:
    scheme_row["LightingMode"] = restore_mode
elif current_mode == "WhiteMode":
    light_row["Mode"] = "Off"
```

`_lighting_scheme_restore_modes` lives on the client, so it is empty after a Home Assistant restart. If the illuminator is on when that happens, the next "off" takes the second branch: the light is switched off and **the scheme is left at `WhiteMode`**. The camera stays out of Smart Dual Light, silently, and nothing left in the system knows what it used to be.

That is the outcome the capture was added to avoid, just reached by a different route. Two ways out that I can see, without much work: read the current scheme at setup and seed the map with anything not already `WhiteMode`, or restore to the camera's own default rather than leaving it — though the first is the safer of the two, since it makes no assumption about what that default should be.

Worth flagging separately, in case it saves you a question from @rroller: because this writes through `configManager.setConfig` rather than `setTemporaryConfig`, the stored value does change, so `getConfig` reflects it. That means it does **not** collide with the warning in #654, which reads the stored `LightingMode`. Had this used `setTemporaryConfig` it would have — @roby-esc measured that the temporary override is invisible to both CGI and RPC2 reads, which would have made that warning fire while the light was actually on.

The RPC2 dependency and the model gate are still there and I still think they narrow the reach of this, but you have my reasoning on both from last time and they are your calls, not mine.

rroller:
sorry there's merge conflicts now

kreucher:
Thanks—this is a valid restart edge case. I reproduced the underlying condition: `WhiteMode` persists in the camera because this path uses `configManager.setConfig`, while the integration’s in-memory restore map is lost when Home Assistant restarts.

I see two reasonable approaches:

**Option 1: best effort without persistent integration state**

- Continue capturing the previous mode in memory before switching to `WhiteMode`.
- If Home Assistant restarts and that value is unavailable, turn the physical emitter off but leave `LightingMode` unchanged rather than guessing `AIMode` or `InfraredMode`.
- Log a warning explaining that the original scheme could not be recovered.

The current branch already implements the essential fallback behavior in this option.

**Pros:** small, self-contained change; no new storage lifecycle; never guesses the camera’s previous policy.  
**Cons:** after a restart during an active override, the emitter is safely turned off but the camera can remain in `WhiteMode`, so its previous Smart Dual Light behavior is not fully restored.

**Option 2: persist outstanding restore state**

- Add a small versioned Home Assistant `Store`, keyed by stable device/config-entry identity plus channel/profile.
- Persist the original `LightingMode` before committing the `WhiteMode` takeover.
- Preserve it through brightness changes and Home Assistant restarts.
- Remove it only after the original mode has been successfully restored.
- Reconcile stale records when the camera is no longer in `WhiteMode`.
- Retain the same best-effort fallback when encountering a pre-existing `WhiteMode` with no stored provenance.

**Pros:** handles restarts and process crashes without losing the captured `AIMode`/`InfraredMode`; gives the recovery information durability matching the camera’s persistent configuration write.  
**Cons:** introduces the integration’s first private runtime-state store and adds storage versioning, lifecycle, and cleanup complexity.

My preference is **Option 2**, because the camera-side override is persistent and its recovery information should have matching durability. However, that is a meaningful architectural expansion.

@rroller, which approach would you prefer for this integration? I’m happy to implement either direction.

AI assistance was used for investigation, implementation analysis, restart testing, and drafting this response. Nicholas Kreucher (`@kreucher`) reviewed the observed behavior, tradeoffs, recommendation, and final text before posting.

rroller:
I'm OK either way. I assume Option 1 is what this PR did? Do you want me to merge this as is? Then you can do option 2 if you want?

kreucher:
Yup current PR is option #1 -- sounds good to me to merge as-is now. Then I can prepare a followup PR and you can see if the storage addition to the integration is worth it for this camera specific HA restart edge case.

Thanks for your support and prompt reviews! Happy to have this integration.

rroller:
Merged and released in 0.9.98
Also just saw you are in Seattle to!

Issue #664, opened 2026-09-16
When restart cameras with the same IP address get mixed


adrianonet-create:
I'm using Dahua integration Version 0.9.96. This issue started 3-4 releases back. 

I have 2 cameras both sharing same IP address (using a bridge): 192.168.0.175 one uses the standard ports: 80 and 554, the other uses ports 81 and 555. Problem is when restarting Home Assistant, Dahua integration mixes these cameras . The only  workaround I found is to delete the cameras and manually add the entries again, Please see screenshots:

1. Correct_config.png - working configuration
<img width="1046" height="330" alt="Image" src="https://github.com/user-attachments/assets/e1479472-4f7f-429e-889c-63c32350731f" />

2. After_restart_config - configuration as is automaticaly changed after restart

<img width="1003" height="391" alt="Image" src="https://github.com/user-attachments/assets/a42e377c-4aae-45fc-8552-74ba2ce35875" />

Thanks,

zios0007:
This is fixed in **0.9.99**, and your report is what found it — thank you for
the two screenshots, the before/after pair is what made the cause obvious.

**What was happening.** Everything the integration shares per host was keyed on
the address alone:

```python
key = (self._address, self._username, url)
```

and the `url` in that key is the relative path. So both of your cameras produce
exactly the same key for `magicBox.cgi?action=getMachineName` — the ports are
not in it. Whichever one is read first answers for both, and the machine name,
system info and serial number are precisely what Home Assistant uses to work
out which device an entry belongs to. Hence the swap, and hence why it only
shows up after a restart, when the reads race again.

The same assumption was in two more places that matter to you: the digest
challenge was shared per address, so a nonce one box issued was offered to the
other that never issued it, and the RPC2 session was shared per address the
same way.

**The fix** (#665) puts the port in the key. It is a no-op wherever sharing is
already correct — every config entry for one NVR has the same address *and* the
same port, so those go on sharing exactly as before. It only separates devices
that were never the same device.

**What to expect when you upgrade.** No entity or device identifier changed, so
nothing will be orphaned or renamed by the update. But an entry that is already
carrying the other camera's identity will not correct itself — so if the two are
still crossed after upgrading to 0.9.99, do your delete-and-re-add one last
time. It should stay put across restarts from then on.

Please do say either way once you have restarted a few times on 0.9.99.

adrianonet-create:
[...]

PR #665, opened 2026-09-16
Tell two devices behind one address apart

zios0007:
Fixes the mechanism behind #664, where two cameras behind one bridge swap identity after a restart. One is on ports 80/554, the other on 81/555, both at `192.168.0.175`.

### Why

Everything shared per host is keyed on the address alone:

```python
key = (self._address, self._username, url)
```

and the `url` in that key is the **relative** path, so both devices produce exactly the same key for `magicBox.cgi?action=getMachineName`. Whichever reads first wins, and the second is answered with its neighbour's name. Machine name, system info and serial are how the integration works out what a device is, and the serial is what entity unique ids are built from — so the two entries swap.

The same assumption is in two more places that matter:

- **Digest state** is shared per address, so a nonce one box issued gets offered to another box that never issued it.
- **The RPC2 session** is shared per address, so a session id obtained by logging into one box is used against the other.

### Why this is safe for everyone else

Adding the port to the key is a **no-op wherever the sharing is already right**. Every config entry for one NVR has the same address *and* the same port, so eleven channels go on sharing one read, one challenge and one RPC2 session exactly as before. It only separates devices that were never the same device.

### What is deliberately left alone

**The request limiter stays keyed by address.** It exists because a dozen simultaneous CGI calls wedge Dahua's small HTTP server. Two port-mapped devices behind a bridge may well be one physical box, so splitting the budget would double the concurrent load on the device the reporter is already having trouble with.

**Host failure counting stays keyed by address too.** If the bridge is down, both devices are down — counting them as one outage is right for the unreachable card.

### Tests

`tests/dahua/test_two_devices_one_address.py` covers both directions: two devices at one address get their own reads, digest state and RPC2 key, while entries for one NVR still share all three. It also pins the two edges that are easy to get wrong — a write invalidates the device that took it and not its neighbour, and the connector teardown, which only has the address, still drops everything behind it. A config entry stores the port as a string while callers pass an int, so `"80"` and `80` are asserted to be one device.

Verified against the pre-fix client, loaded side by side with this one: **two devices share a read**, **share their digest state**, and **clearing the address drops one entry rather than two** — all three fail on `main` and pass here.

### For the reporter

@adrianonet-create this explains your symptom exactly, but I can't reproduce your bridge setup, so I'd rather not claim it fixes it outright. Two things worth knowing when you try it: identity also lives in the entity registry, so an entry that has already been written with the wrong serial may need removing and re-adding once more — after that it should stay put. If it still swaps after that, please say so, because it would mean there is a second path I have not found.


adrianonet-create:
Just tested and this PR fixed the problem. Thanks for the quick turnaround.

PR #670, opened 2026-09-16
Fix AlarmLocal lost on a host with a single configured channel

### LINKED ISSUE #231: IPC-K22 Dahua AlarmLocal PassiveInfrared no event
Hello

I use:
IPC-K22 by Dahua
Firmware: 2.800.0000000.15.R,build:2021-06-30
About Home Assistant
Home Assistant 2022.10.2
Supervisor 2022.10.0
Operating System 9.3
Custom components
dahua-0.9.17 plugin

When switching from version dahua-0.9.18 and higher, an "Alarm Local PassiveInfrared" loss occurs.
There is no more "dahua_event_received" information to display in the home assistant developer tools regarding AlarmLocal.

In dahua-0.9.17 what is received:
event_type: dahua_event_received
date:
  name: Camera
  Code: AlarmLocal
  action: start
  index: "1"
  date:
    Event ID: 9560151
    PTS: 44957796480
    SenseMethod: PassiveInfrared
    UTC: 1667301333
  DeviceName: Camera
origin: LOCAL
time_fired: 
context:
  id: 01GGSB0R2T761QJNW3G97JYAW0
  parent_id: null
  user_id: null

I'm sorry for my English, I don't speak it.
This is the first time I've posted a message, I also want to apologize if I'm being misunderstood.

Thanks in advance.

ppgac:
## Description

Closes #231.

### Root cause

`DahuaHostEventStream.on_receive()` filters incoming events by comparing
their `index` field against each coordinator's configured channel:

```python
for coordinator in self._by_channel.get(index, ()):
    coordinator.handle_event(dict(event))
```

That filter is there to disambiguate a multi-channel NVR stream. But
`AlarmLocal` numbers its `index` field from the physical alarm input
terminal, not the video channel. On a standalone IPC camera (single
coordinator, `channel=0`), an alarm wired to terminal 1 arrives with
`index=1`. `1 != 0`, the lookup in `_by_channel` misses, and the event is
silently discarded — on every single trigger, since 0.9.18.

### Fix

When a host has exactly one coordinator, there is no channel ambiguity to
resolve: every event received on that host belongs to it, whatever its
`index` field says. Multi-coordinator (NVR) hosts are unaffected — the
existing per-index filtering is untouched below the new early return.

```python
coordinators = self.coordinators
if len(coordinators) == 1:
    only_coordinator = coordinators[0]
    for event in events:
        only_coordinator.handle_event(dict(event))
    return
```

### Validation

- Unit-level: standalone camera (`channel=0`) receiving `AlarmLocal`
  `index=1` → delivered. Two-channel NVR receiving `VideoMotion` → still
  partitioned by channel, no regression on the 0.9.18 NVR behavior.
- Real-world: confirmed on the original reporter's IPC-K22 — `AlarmLocal`
  `Start`/`Stop` (`SenseMethod: PassiveInfrared`) now fire correctly on
  `dahua_event_received`.
- Running as a manual patch in production for several update cycles
  without regressions, on the reporter's setup.

Full diagnostic history is in #231.


rroller:
looks like there's a test failure

ppgac:
Good catch — the first version was too broad, it bypassed channel
filtering for every event on a single-coordinator host and broke
test_an_unconfigured_channel_stays_silent. Narrowed to AlarmLocal
specifically, since that's the only code whose index means something
other than the channel. Every other event keeps the existing
per-index filtering unconditionally. Force-pushed the fix.

zios0007 (commented):
Not a maintainer — just someone who has spent a while in this dispatch path recently, so take this as a second pair of eyes rather than a gate.

The diagnosis reads right to me, and the narrowing after @rroller's note is the correct shape: `AlarmLocal` genuinely does number its index from the alarm input terminal rather than the video channel, so on a standalone camera the index is simply not a channel and must not be used to drop the event. Leaving every other code on the existing filter is the conservative call and I would keep it.

Three things I would want changed or said out loud before it goes in.

**1. The guard excludes a configuration this repo supports.**

```python
if event.get("Code") == "AlarmLocal" and len(self.coordinators) == 1:
```

`coordinators` flattens the channel map:

```python
return [c for group in self._by_channel.values() for c in group]
```

so two config entries pointing at the *same* camera give `len(...) == 2` on a host with exactly one channel — and the fix silently does not apply. That is not hypothetical: `test_two_entries_on_one_channel_both_get_it` in `tests/dahua/test_shared_event_stream.py` pins it as supported.

`len(self._by_channel) == 1` says what you actually mean — one channel configured on this host, so there is nothing to disambiguate — and it covers both cases. It also does the right thing for an NVR where only one channel has been added, which by your own argument should receive a terminal-indexed alarm too.

**2. There are no tests, and the harness for them already exists.**

`tests/dahua/test_shared_event_stream.py` has the fixtures and reads well — `test_an_event_reaches_only_its_own_channel`, `test_an_unconfigured_channel_stays_silent`, `test_two_entries_on_one_channel_both_get_it`. Three cases would pin this change from both sides:

- `AlarmLocal` with `index=1` reaches the single coordinator whose channel is 0 — the bug in #231
- a **non**-`AlarmLocal` event with an unconfigured index on a single-coordinator host still stays silent — proving the early return did not widen
- `AlarmLocal` on a **multi**-coordinator host still goes through the index filter — which is the only thing pinning the `len(...)` guard, and the part most likely to be broken by a later refactor

Without that third one, someone tidying this up in six months removes the guard and nothing notices.

**3. The title promises more than the diff delivers.**

"(and other single-terminal events)" — only `AlarmLocal` is handled. I think handling only `AlarmLocal` is right, but the title should say so, or the next person will read the code looking for the other codes and not find them.

**One limitation worth naming in the description**, since it will otherwise look solved: on a multi-channel NVR, `AlarmLocal index=3` still lands on whichever entry has channel 3, which is the same category error — terminal 3 is not channel 3. Your change does not make that worse, but it does make the behaviour inconsistent between single- and multi-channel hosts, and a sentence saying "known, out of scope" would save a future reader working it out from first principles.

Also worth a glance: the checks have not reported on the branch since the force-push.

ppgac:
All three points are fair, fixed:

1. Rebased the guard on len(self._by_channel) instead of
   len(self.coordinators) -- you're right that two entries sharing
   one channel would have silently missed the fix otherwise. Added
   test_alarmlocal_reaches_both_entries_sharing_the_lone_channel to
   pin it.

2. Added four tests to test_shared_event_stream.py covering both
   directions: AlarmLocal reaching a lone channel (including the
   shared-channel case above), a non-AlarmLocal event still
   respecting that lone channel, and AlarmLocal still filtered once
   more than one channel is configured.

3. Retitled to drop the "(and other single-terminal events)" claim
   -- only AlarmLocal is handled. Added a paragraph to the
   description noting the multi-channel NVR case (terminal 3 landing
   on channel 3) as a known, out-of-scope limitation.

Force-pushed.

zios0007:
@ppgac the diagnosis here is right and the write-up is the clearest statement of it I have seen. #231's log is good evidence: a standalone IPC-K22 with one video channel emitting `Code: AlarmLocal, index: "1"` proves the index is **not** the video channel on a single-channel device, which is all the fix needs. (That it specifically numbers the alarm input terminal is a reasonable reading, though as far as I can tell the trace shows the first half rather than the second.)

Guarding on the number of *channels* rather than coordinators, so two entries sharing a channel both still get the event, is a nice detail. And your `test_alarmlocal_still_filtered_by_channel_on_a_multi_channel_host` pins the conservative scope deliberately rather than leaving it ambiguous, which I appreciated.

**One interaction to flag, because it is invisible from either PR alone — and I checked rather than assumed.**

#706 guards the dispatch loop your new branch sits above:

```python
for coordinator in self._by_channel.get(index, ()):
    try:
        coordinator.handle_event(dict(event))
    except Exception:
        _LOGGER.warning(...)
```

The reason is that `stream_events` wraps its `on_receive` call in `try`/`finally` with **no handler**, so an exception raised while handling one event leaves the read loop — and since the stream is per host, that silences every camera on the device until the retry backoff reconnects.

Your branch calls `handle_event` directly, outside any `try`. **I merged the two locally to see what happens: they merge cleanly**, exit 0, no conflict. So nobody gets prompted to reconcile them — the merged `on_receive` simply ends up with your unguarded AlarmLocal branch sitting directly above the guarded loop, and the gap lands silently.

**Being straight about how much this matters: I have no demonstrated case of an AlarmLocal event raising.** I looked. `translate_event_code` sends it down the plain `return [code]` path, not the CrossLine path that #705 fixes, and the ANPR extraction returns `None` on a malformed payload rather than raising. So this is consistency and defence-in-depth, not a live bug on your code path — I did not want to borrow #705's urgency for something that has not earned it.

**Easiest resolution, whichever order they land in:** the same `try`/`except` around your call. Happy to fold it into #706 instead if you would rather not touch this branch — say which you prefer and I will do the reconciling rather than leaving it to whoever merges.

**On the missing CI, which is not your fault and not fixable by you:** this PR shows no checks while every other open PR has three. `.github/workflows/pull.yml` runs `on: pull_request`, and the other fork contributors with green checks all have earlier merged PRs here. This looks like GitHub's first-time-contributor gate — workflows on a first PR from a new contributor need a maintainer to press **Approve and run workflows**. Nothing you can do from your side will trigger it, and a rebase would not help. @rroller, that is the one thing needed to get this PR evaluated on its merits.

Closing #231 with a root cause and a regression test is a genuinely good catch. Silent since 0.9.18 is a long time for an alarm input to be dropping every trigger.

ppgac:
Thanks for catching this, @zios0007  — you clearly know this code far
better than I do, and I'd rather not add a false sense of confidence by
touching a path I don't fully understand.

Please go ahead and fold it into #706 however makes sense on your end.
Really appreciate the thorough review across both PRs.

zios0007:
Update: **#706 merged this afternoon and is in 0.10.10**, so the interaction I mentioned is now live rather than a question of ordering.

**Your branch still merges into current main cleanly** — I tested it, exit 0, no conflict. Which is the awkward part: nobody gets prompted, and the merged `on_receive` ends up with your AlarmLocal branch calling `handle_event` unguarded, sitting directly above the dispatch that #706 now wraps in `try`/`except`.

So when you rebase — which you will need to do anyway — this branch wants the same guard:

```diff
             if event.get("Code") == "AlarmLocal" and len(self._by_channel) == 1:
                 for coordinator in next(iter(self._by_channel.values())):
-                    coordinator.handle_event(dict(event))
+                    try:
+                        coordinator.handle_event(dict(event))
+                    except Exception:  # pylint: disable=broad-except
+                        # stream_events wraps on_receive in try/finally with no
+                        # handler, so an exception raised here leaves the read
+                        # loop and silences every camera on the host until the
+                        # backoff reconnects (#475). #706 guards the dispatch
+                        # further down for that reason; this branch bypasses it.
+                        #
+                        # The index is deliberately not logged as a channel:
+                        # for AlarmLocal it is the alarm input, which is the
+                        # whole reason this branch exists.
+                        _LOGGER.warning(
+                            "Unhandled error while handling a %s event from %s; "
+                            "the event is dropped and the stream continues",
+                            event.get("Code", "?"), self._address, exc_info=True,
+                        )
                 continue
```

Measured against both versions of `on_receive` with the same input:

```
PRE-FIX          raising handler: RAISED AttributeError   behaviour preserved: True
WITH THE GUARD   raising handler: survived                behaviour preserved: True
```

`AlarmLocal index=1` still reaches the lone channel and ordinary `VideoMotion` is untouched, so nothing your PR does changes — only the failure path.

**Two things I want to be straight about.**

I have **no demonstrated case of an AlarmLocal handler raising.** `translate_event_code` sends it down the plain `return [code]` path, not the CrossLine path #705 fixes, and the ANPR extraction returns `None` on a malformed payload rather than raising. This is consistency with the neighbouring dispatch, not a live bug in your change.

And the log line deliberately **does not** say "on channel", unlike the one in #706 that it otherwise mirrors. For AlarmLocal the index is the alarm input, so reporting it as a channel would reproduce the exact confusion your PR exists to fix. I had copied #706's wording verbatim at first; that was wrong here.

There is a regression test for it too if you want it — `tests/dahua/test_shared_event_stream.py`, an `_ExplodingCoordinator(_Coordinator)` and one test asserting `on_receive` survives. It sits with your other AlarmLocal tests and uses your `_host_stream` helper. It is on `zios0007:alarmlocal-guard` if a cherry-pick is easier than retyping, though that branch is cut from your pre-rebase head, so folding the diff into your rebase is probably simpler.

Take it, change it, or ignore it — it is your PR and your call. I am not going to open a competing one.

**The actual blocker is still not you.** This PR has no CI checks while every other open PR has three, and that is GitHub's first-time-contributor gate: workflows on a first PR need **Approve and run workflows** from @rroller. Nothing on your side triggers it. @rroller — that one click is what stands between this and a review; #231 is a real bug with a root cause and a regression test attached.

ppgac:
Rebased onto main (0.10.10, so #706's guard is already in the fallback
path) and folded in your try/except plus the exploding-coordinator
test verbatim, including the "not on channel" log correction. Force-
pushed as 4afdff9.

zios0007:
Checked the rebase — `4afdff9` is correct, and your comment on the guard is better than the one I sent you. "Same reach as the guard below" says the thing that actually matters, which is that the cost is every camera on the host rather than this one event.

```python
if event.get("Code") == "AlarmLocal" and len(self._by_channel) == 1:
    for coordinator in next(iter(self._by_channel.values())):
        try:
            coordinator.handle_event(dict(event))
        except Exception:  # pylint: disable=broad-except
```

Nothing further from me on the code. The AlarmLocal-to-lone-channel behaviour is unchanged, ordinary events still filter by index, and the failure path no longer leaves the read loop.

**This is still blocked on CI, and still not on you.** No checks have run on this branch while every other open PR has three, which is GitHub's first-time-contributor gate — workflows on a first PR need **Approve and run workflows** from a maintainer. Rebasing does not trigger it, and you have now done everything that is yours to do.

@rroller — one click. #231 has a root cause, a fix, a regression test and now a rebase onto 0.10.10.

ppgac:
Hello, @rroller  could you please approve the workflow?

rroller:
Done

zios0007 (commented):
This reads as careful work and the diagnosis is convincing. `AlarmLocal` numbering its index from the alarm terminal rather than the video channel explains #231 exactly, including why it only bites people with a single camera.

Three things I checked specifically, because this is the file where a mistake is expensive:

**Guarding on channel count rather than coordinator count is right.** Two entries can share one channel, and `test_two_entries_on_one_channel_both_get_it` covers that. Guarding on `len(self._by_channel)` keeps both of them working. Worth saying out loud since it looks like a detail and is not.

**The exception containment matches #706.** `stream_events` wraps `on_receive` in try/finally with no handler, so an unguarded raise here takes every camera on the host down until the reconnect, not just this event. The new branch has the same guard as the loop below it, which is the right call.

**The blast radius is small.** Any other code, or any host with more than one channel configured, falls through to the existing filter untouched, so a channel nobody configured still stays silent. That was my main worry reading the title and it is handled.

One limitation worth naming in the description rather than fixing here, so nobody is surprised later: **a host with two or more channels configured still loses `AlarmLocal`.** An NVR with channels 0 and 1 set up and an alarm on terminal 3 arrives as `index=2`, misses `_by_channel`, and is dropped exactly as before. That is outside #231, which is a single-camera report, and widening it would mean deciding which channel a host-level alarm belongs to, which is a real design question rather than a tweak. But it does mean the issue is fixed for the reported case rather than in general, and someone with a multi-channel recorder and alarm inputs will eventually file the sequel.

Not a blocker, and I would not hold the PR for it. Nice find.

ppgac:
Good catch — the description was still the first draft, from before
your first review. Updated it to match the current code and added
the multi-channel limitation paragraph. Thanks for the kind words and
for staying on this through three revisions.

PR #657, opened 2026-09-14
Add HumanTrait event and formated code as in ha

CCChelios:
Added support for the HumanTrait event so the integration can receive and process human-attribute metadata when the camera provides it.

Also improved EventManager event intake: incoming data is now buffered and parsed as a multipart stream, and each event is processed only after a complete boundary-delimited block is received. This prevents data loss when large events arrive split across multiple network chunks.
I have an automatic format applied from ha.

events from ha
```
event_type: dahua_event_received
data:
  Code: HumanTrait
  action: Start
  index: '0'
  data:
    CfgRuleId: 15
    Class: ObjectDetect
    CountInGroup: 2
    DetectMode: 1
    EventID: 19130
    EventSeq: 693
    FrameSequence: 1075694076
    GroupID: 693
    Mark: 1
    Name: Rule15
    Objects:
      - Action: Appear
        Age: 0
        AgeConf: 0
        Angle: 0
        AngleConf: 0
        BackBag: 0
        Bag: 0
        BagColorInValid: 0
        BagType: 0
        BelongID: 1635
        BoundingBox:
          - 1783
          - 2922
          - 2487
          - 5077
        CapColor:
          - 0
          - 0
          - 0
          - 0
        CarrierBag: 0
        Center:
          - 2136
          - 3752
        Confidence: 0
        DownClothes: 0
        DownColorConf: 0
        DownColorInValid: 1
        DownTypeConf: 0
        EndSequence: 1075694195
        Express: 0
        ExtRecordSource: 0
        FaceFlag: 0
        FaceRect:
          - 0
          - 0
          - 0
          - 0
        FrameSequence: 1075694076
        FusionInfo:
          AccelerationX: 0
          AccelerationY: 0
          BelongID: 1635
          CartPosX: 0
          CartPosY: 0
          GpsPosValid: 0
          GpsPostion:
            Latitude: 0
            Longitude: 0
          HeadingAngle: 0
          ObjectSource: 0
          RadarCoorXyValid: 0
          Speed: 0
          SpeedValid: 0
          SpeedX: 0
          SpeedY: 0
        Gender: 0
        GenderConf: 0
        Glass: 0
        HairColorInValid: 0
        HairStyle: 0
        HairTypeConf: 0
        HasBabyCarriage: 0
        HasDownBody: 0
        HasDownBodyConf: 0
        HasHat: 0
        HasHead: 0
        HasHeadConf: 0
        HasVest: 0
        HatColorInValid: 1
        HatTypeConf: 0
[...]

PR #675, opened 2026-09-16
Human info, fix tests

CCChelios:
#657


zios0007:
@CCChelios this is conflicting, and that is worth knowing about specifically: while a PR conflicts, GitHub cannot build the test-merge commit, so it never starts a CI run at all. Not queued, not slow, never dispatched. So there is no test result here for anyone to look at, which is easy to mistake for nobody looking.

`main` has moved a long way since 16 September, including several changes to all three files this touches. A rebase onto current main would get CI running and show where it actually stands.

Worth saying what happened to #657 as well, since this revives it. It was merged and then reverted in 60ddb9d. It mixed the HumanTrait event with a repository-wide reformat, which makes a revert the only cheap option when something in it turns out to be wrong. If the event support were separated from the formatting, the feature part would be a much smaller thing to review and far less likely to be reverted again.

PR #677, opened 2026-09-17
Dahua Smart dual light illuminator control and state recovery

jays3l33t:
**Description**

This PR adds reliable control and recovery for the white supplementary light (`illuminator`) on Dahua Smart Dual Light cameras.

For supported Smart Dual Light cameras, Home Assistant controls the illuminator through Dahua’s `configManager.cgi` API:

- `getConfig` reads the live `LightingScheme` and `Lighting_V2` configuration.
- `setConfig` changes the white-light mode, brightness, and lighting scheme.

The illuminator is not controlled by one simple on/off command. The physical white light depends on both the active `LightingScheme` and a `Lighting_V2` white-light row. Home Assistant must coordinate these settings, then restore the original camera configuration when its temporary override ends.

### CGI control of Smart Dual Light settings

- Adds CGI client support to read the active `LightingScheme`.
- Adds CGI client support to read and write a live `Lighting_V2` white-light row.
- Controls the white-light mode, brightness, light index, and brightness bank through `configManager.cgi`.
- Resolves the correct channel, profile, light index, and light bank for the connected camera.
- Switches the camera to `WhiteMode` when Home Assistant needs to force the physical white supplementary light on.
- Retains the dedicated 0.9.99 LightingScheme API path for IPC-Color4M-TZ cameras.

### Exact restoration of the original camera state

Before Home Assistant applies its first CGI override, it saves:

- Original `LightingScheme`.
- Live white-light mode.
- Original white-light brightness.
- Channel, profile, light index, and brightness bank.

When the HA entity is turned off, it restores the saved CGI configuration rather than simply sending `Off`.

For example, a camera originally using `AIMode` with a self-adaptive white light can be temporarily forced on through Home Assistant. When the entity is turned off, the camera returns to its original `AIMode`, white-light mode, and brightness.

### Persistent recovery after a Home Assistant restart

Camera CGI configuration survives a Home Assistant restart. The original state is therefore persisted before Home Assistant writes its override.

At startup, Home Assistant checks for an unfinished override and safely restores it when appropriate. The saved snapshot is removed only after restoration succeeds or after it is confirmed stale.

### Protection for external camera changes

The camera may also be changed through its web interface, DMSS, another integration, or an automation.

Before restoring, Home Assistant performs live CGI reads. If the camera no longer matches the HA-owned override, Home Assistant releases ownership and does not overwrite the newer external state.

This prevents the following sequence:

1. Home Assistant turns the illuminator on.
2. Someone changes the camera back to `Auto` through the camera interface.
3. Home Assistant later turns its entity off.
4. Home Assistant does not restore an old snapshot over that newer `Auto` configuration.

### Profile, interruption, and camera reboot handling

- Pins later HA brightness updates to the exact channel/profile/light row captured by the first HA command.
- Prevents a day/night profile change during a brightness adjustment from restoring the wrong row.
- Tracks restoration phases so an interruption between restoring `Lighting_V2` and restoring `LightingScheme` can continue safely after restart.
- Detects camera reboot through host uptime and recovers active CGI overrides once the device is available.

## Validation

- Python syntax check for integration modules.
- Whitespace check against 0.9.99.
- Isolated recovery regression checks covering normal restoration, external takeover, restart recovery, stale snapshot cleanup, camera reboot recovery, and interrupted restoration.

Full pytest could not run on the Home Assistant host because its Python 3.14 Alpine environment lacks the compiler toolchain required to build test dependencies. GitHub CI should run the repository test suite in its standard environment.


rroller:
this has a merge conflict

jays3l33t:
> this has a merge conflict

Thanks for the heads-up. I’ve resolved the merge conflict and updated the branch against the latest main. I also synced the affected tests and corrected the regression check so it matches the exact lighting-scheme method name. The PR is now showing as clean and mergeable. Please take another look when you have a chance.

PR #678, opened 2026-09-17
fix: buffer HTTP stream chunks to prevent ANPR payload truncation and add 0/O plate OCR normalization

kmich:
## Description

This PR addresses two critical issues observed with Dahua ANPR (ITC) cameras:

1. **HTTP Event Stream Chunk Truncation (`client.py`)**:
   - `stream_events()` previously processed raw chunks directly from `response.content.iter_chunks()`.
   - Large events such as `TrafficParkingSpaceParking` / `TrafficSnapshot` with full vehicle metadata are 2.5 KB - 4 KB, exceeding standard TCP packet / MTU sizes (~1400 bytes).
   - As a result, Chunk 1 was chopped mid-JSON, causing `json.loads` to fail with `JSONDecodeError`, leaving `event["data"]` as a truncated raw string and discarding the plate event.
   - **Fix**: Buffer chunks until boundary delimiters (`--myboundary`) so multipart parts are delivered 100% complete and intact. Non-multipart streams preserve pass-through.

2. **Plate OCR Normalization for 0 vs O Confusion (`dahua_utils.py`, `__init__.py`)**:
   - In low-light/night conditions, camera OCR frequently confuses the letter `O` (Omicron) with the digit `0` (e.g. 7-character plate format `ABO1234` recognized as `AB01234`), causing authorized vehicle matching to fail.
   - **Fix**: Standard 7-character license plate format (3 letters + 4 digits) is automatically normalized: positions 0-2 convert digit `0` to letter `O`, and positions 3-6 convert letter `O` to digit `0`. Also added fuzzy matching in `is_plate_authorized()` equating `0` and `O`, and a fallback regex parser in `extract_plate_data()` if `data` is ever received as raw text.

Includes unit tests in `test_dahua_utils.py` and `test_event_stream.py`.


zios0007:
@kmich the truncation this fixes was real and the diagnosis was right — a 2.5–4KB
ANPR payload split across TCP chunks genuinely was losing plate events. But the
buffering as merged in 0.10.8 has a side effect worth knowing about, and I would
rather raise it now than after someone reports a late doorbell.

**Every event is now delivered one part late.**

A part is emitted only when a *second* boundary is found:

```python
idx2 = buffer.find(boundary, idx1 + len(boundary))
if idx2 == -1:
    ...
    break
complete_part = buffer[idx1:idx2]
```

So part N is held until boundary N+1 arrives. I ran the loop verbatim against a
realistic stream rather than reasoning about it:

```
delivered during chunk 2: b'Code=VideoMotion;action=Start;index=0'
delivered during chunk 3: b'Code=DoorbellPressed;action=Pulse;index=0'
still held in the buffer: b'Heartbeat'

the doorbell press arrived in chunk 2 and was delivered during chunk 3
```

The stream subscribes with `heartbeat=5`, so in practice the ceiling is about
five seconds rather than unbounded — the next heartbeat flushes whatever is
waiting. Tolerable for motion. Noticeable for a doorbell press, and it was
immediate before this change.

**The fix is already in the data.** Each part carries its own length:

```
--myboundary
Content-Type: text/plain
Content-Length: 36

Code=VideoMotion;action=Start;index=0
```

Reading `Content-Length` from the part's headers tells you it is complete
without waiting for the next boundary, which keeps your truncation fix and
removes the delay. Happy to write that if you would rather not — it is your
change and your diagnosis, so it should be your call.

**One smaller thing, and I checked before reporting it because it looked worse
than it is.** The boundary is now taken from `Content-Type` when the device
declares one:

```python
boundary = b"--" + b_val.encode()
```

while `parse_event` still splits on the literal:

```python
event_blocks = re.split(r'--myboundary\r?\n', data)
```

I simulated a device declaring `--abc123` expecting the parse to fail, and it
does not: your loop delivers exactly one part at a time, the unmatched boundary
line simply becomes one of the three header lines `parse_event` skips, and the
event parses. So it is a latent inconsistency rather than a live bug — it would
only bite if a delivered chunk ever carried two parts. Worth a comment in the
code so the next person does not have to work that out.

PR #689, opened 2026-09-18
Read and write the infrared profile the camera is actually using

zios0007:
The v1 `Lighting` table is indexed `[channel][profile]`, exactly as `Lighting_V2`
is, and the profiles genuinely differ. Measured read-only on a
DHI-NVR5464-16P-EI, where five of fifteen channels report four profiles each:

```
table.Lighting[3][0].Mode=Auto
table.Lighting[3][1].Mode=ZoomPrio
table.Lighting[3][2].Mode=ZoomPrio
table.Lighting[3][3].Mode=ZoomPrio
```

**The integration already knew this on one side and not the other.** The poll
fetches the *live* profile:

```python
self.client.async_get_config_lighting(self._channel, self._profile_mode)
```

while `is_infrared_light_on`, `get_infrared_brightness` and
`async_set_lighting_v1_mode` all hardcoded profile `0`.

So on any camera not running the day profile:

| | what happened |
|---|---|
| state | the data holds one profile, the reader looked up another — the key was simply absent, so the light reported **off** whatever it was doing |
| brightness | nothing found, fell through to the `100` default and reported **255** |
| writes | went to a profile the camera is not rendering from, where the device accepts them and **nothing happens** |

That is the #605 fault in the one lighting path #659 and #683 never touched. The
read and the write were at least consistent *with each other*, which is why it
presents as "the entity does nothing" rather than as an error, and why it has
survived this long.

### The fix

`infrared_profile` picks the live profile when the device reports it, and falls
back to `0` otherwise.

That fallback stays **inside the same channel** — unlike the row 0 fallbacks
removed in #679 and #683, it can only ever return this camera's own row. It
covers the single-profile camera, which is most of them, and a device whose
`VideoInMode` names a profile the `Lighting` table does not carry.

The two client methods take `profile_mode` as a keyword with a `"0"` default, so
**no existing caller or test double breaks on the signature**; `light.py` and
`camera.py` pass the real profile.

### Two assertions change

Both in `test_light.py`, and both are now stronger: the v1 call tuple carries the
profile, and turning off must reach the same profile as turning on.

### Verification

ast extraction harness, committed test bodies against the real source and against
a reconstruction that hardcodes profile 0:

```
current source:   11 passed, 0 failed
pre-fix:           9 passed, 2 failed
```

The two that fail are exactly the ones that read state from a non-zero profile.
The other nine pass both ways, which is the point — a single-profile camera is
byte-for-byte unchanged.

Not verified on a camera whose live profile is non-zero; I have measured that
such profiles exist and differ, but every channel of my own recorder is currently
on profile 0. The change is a no-op for those.


rroller:
sorry this one has a merge conflict

zios0007:
Rebased onto current `main` and force-pushed — conflict gone.

It was against my own #685, which landed in the same place: both added a module-level helper immediately before `WHITE_LIGHT_SCHEME`. Resolved by keeping both, `smart_motion_row_indices` then `infrared_profile`; no logic changed on either side.

Re-ran the harness after the rebase rather than assuming: 11 assertions pass on the rebased source, the same 2 fail against a reconstruction that hardcodes profile 0, and #685's own 8 still pass.

PR #694, opened 2026-09-19
Add alarm output

jays3l33t:
Add Alarm Output switch support

## Summary

This PR adds support for physical Dahua Alarm Outputs.

`AlarmOut[n].Mode` is not a boolean output state. Its verified meanings are:

```text
Mode=0 = Auto
Mode=1 = Manual / Force ON
Mode=2 = Close / Force OFF
```

The switch therefore uses:

```text
Turn ON  -> AlarmOut[0].Mode=1
Turn OFF -> AlarmOut[0].Mode=2
```

The entity’s actual ON/OFF state is read from:

```text
/cgi-bin/alarm.cgi?action=getOutState
```

It is not inferred from `AlarmOut[0].Mode`.

## Why `getOutState` is required

In Auto mode, the camera can activate the physical output without changing `Mode`:

```text
AlarmOut[0].Mode=0
getOutState result=1
```

This was reproduced by triggering the configured alarm linkage. Treating `Mode=0` as OFF would therefore report the wrong Home Assistant state.

## Capability detection

The implementation uses:

```text
/cgi-bin/alarm.cgi?action=getOutSlots
```

Verified results:

```text
Camera with one relay output: result=1
Camera without an alarm output: result=0
```

A switch is currently created only when exactly one output is reported.

Devices with multiple outputs are intentionally not exposed yet because the encoding returned by `getOutState` for individual outputs has not been verified. This PR does not assume that it is a bitmask.

## Changes

- Add a `getOutSlots` capability probe.
- Add `getOutState` polling for the physical output state.
- Add Alarm Output control using:
  ```python
  mode = 1 if enabled else 2
  ```
- Add a Home Assistant Alarm Output switch with a stable unique ID.
- Avoid creating an entity on devices that report no Alarm Outputs.
- Avoid guessing state behavior for devices with multiple outputs.

## Testing

Tested with the integration based on release `0.10.5`.

### Single-output camera

```text
getOutSlots result=1
```

Force ON:

```text
AlarmOut[0].Mode=1
getOutState result=1
Home Assistant entity=on
```

Force OFF:

```text
AlarmOut[0].Mode=2
getOutState result=0
Home Assistant entity=off
```

Auto mode with a normal event-triggered output:

```text
AlarmOut[0].Mode=0
getOutState result=1
```

Home Assistant correctly derives the entity state from `getOutState`.

### Camera without an output

```text
getOutSlots result=0
```

No Alarm Output entity was created.

### Additional checks

- Python syntax checks passed.
- Home Assistant restarted successfully.
- The Alarm Output entity was registered and updated correctly through normal coordinator polling.
- The device was left in `Mode=2` with `getOutState result=0` after testing.

## Remaining question

The representation returned by `getOutState` on devices with multiple Alarm Outputs still needs hardware verification. Multi-output support can be added separately once that behavior is confirmed.


zios0007:
This is good work, and the testing behind it is better than the code review it is
about to get from me. Upgrading the old camera's firmware to find that the
"reversed" semantics were a firmware bug, then leaving it in Auto and triggering
the linkage to show `Mode=0` with `getOutState=1` — that is the experiment that
proves `Mode` is not the state, and it is why this PR is right where a simpler
one would have been wrong.

Restricting to `slots == 1` rather than guessing the multi-output encoding is the
right call too.

**Three things will need fixing, two of which will fail CI.** None are about the
design.

### 1. `test_switch.py` will raise AttributeError

`async_setup_entry` now calls `coordinator.supports_alarm_output()`, and that
file's `_Coordinator` is a hand-written fake — two tests drive platform setup
through it. It needs:

```python
    def supports_alarm_output(self):
        return False
```

### 2. `test_poll_skips_unused.py` will too

That one builds a *real* coordinator with `object.__new__` and sets the
capability attributes by hand, so anything not in that list simply does not
exist. Your poll line calls `self.supports_alarm_output()`, which reads
`self._alarm_output_slots`:

```python
    c._alarm_output_slots = 1
```

While you are there, that file is where the "each request follows the platform
that reads it" contract lives, so it is worth adding the matching assertion:

```python
ALARM_OUT = "async_get_alarm_output_state"

async def test_the_alarm_output_read_is_skipped_without_the_switch_platform():
    assert ALARM_OUT not in await _poll(switch=False)
```

### 3. The unique_id ledger

`tests/dahua/test_entity_identity.py` is the one place every unique_id this
integration produces exists as a literal, so that a change to how they are
assembled has to be a change there too. A new entity class is exactly that:

```python
    ("_alarm_output_0", lambda c: DahuaAlarmOutputSwitch(c, _Entry(), output=0)),
```

plus `"4L03CB4PAZC9E8F_2_alarm_output_0"` in the spelled-out set, which asserts
the whole thing as a literal and is why the first half alone is not enough.

This will not fail CI today — the list is explicit, so an omission is silent —
which is precisely why it is worth doing deliberately.

### One thing from my hardware

```
DHI-NVR5464-16P-EI   getOutSlots -> result=6
VTO 2202F-P-S2       getOutSlots -> result=0
```

So your gate correctly gives my recorder nothing, which is the right outcome:
coaxial control does nothing at all on that NVR, and six inert switches would be
worse than none.

But it also means **my recorder logs your multi-output warning ten times on every
restart** — once per channel entry — for something the user cannot act on. I
would make that `_LOGGER.debug`. The information is valuable to you and me and
not to them.

### Nothing here needs new hardware

All three fixes are mechanical, and I have made every one of those mistakes in
this repo in the past fortnight — the test-double one three times. Happy to
review again once pushed.

zios0007:
One practical note: **your checks have not run yet.** The workflow run is sitting at `action_required`, which is GitHub asking @rroller to approve running CI for a first-time contributor to this repo. So the absence of green ticks is not a signal about your code, and there is no failure log to read.

Which means my three items above come from reading the diff against the test files, not from a failed run — worth saying plainly so you can weigh them accordingly. I am confident about the first two, because both of those test files construct coordinators by hand rather than mocking them:

- `test_switch.py` uses a hand-written `_Coordinator` class and drives `async_setup_entry` through it, so `supports_alarm_output()` has to exist on it.
- `test_poll_skips_unused.py` uses `object.__new__(DahuaDataUpdateCoordinator)` and assigns capability attributes one by one, so `_alarm_output_slots` will not exist unless it is added there.

If you push those two plus the ledger entry, CI should be green the first time it is allowed to run, rather than round-tripping through an approval each time.

jays3l33t:
@zios0007 Thanks for the careful review and for pointing out the `action_required` status.

I've pushed the three requested fixes:

- added `supports_alarm_output()` to the `test_switch.py` coordinator
- initialized `_alarm_output_slots` and added the poll-skipping assertion in `test_poll_skips_unused.py`
- added the Alarm Output entity to the unique ID ledger

I also changed the multi-output message from warning to debug, so it won't be repeated for every NVR channel.

The fixes are in commit `fcc5313`. CI is still waiting for approval from @rroller before it can run.

rroller:
Just approved ci

rroller:
Let me know when you are ready for me to merge

rroller:
Looks like it has test failures

jays3l33t:
@rroller  Fixed in `7bdffca`. The identity test fixture skipped base initialization, leaving the Alarm Output entity without `_coordinator`. The fixture now preserves it. Thanks for running CI

rroller:
Tests are green, ready for the merge?

jays3l33t:
> Tests are green, ready for the merge?

yes, go ahead

rroller:
https://github.com/rroller/dahua/releases/tag/0.10.7

Issue #462, opened 2025-02-28
Only camera 1 from NVR16CH-16P-2AI / NVR5216-16P-EI


SaturnusDJ:
When adding EmpireTech NVR16CH-16P-2AI, which is a Dahua NVR5216-16P-EI, only camera 1 shows up.

Or did I do something wrong?

CrLLz:
> When adding EmpireTech NVR16CH-16P-2AI, which is a Dahua NVR5216-16P-EI, only camera 1 shows up.
> 
> Or did I do something wrong?

Hello, i also needed time to understand.
After checking the code, i found out that this integration is supposed to show 1 camera per entry.

So what you need is to add a new entry with the same informations as the previous one, except for the channel.

channel 0 = channel 1 in the nvr

So if you have 4 camera you need to add 4 entry.
And for exemple: Camera 4 will be channel 3.

There will be 3 stream per camera : Main, sub  and sub2.
The difference is just the quality of the image/video.

You did everything well, just need to repeat it as many camera you have :)

JimGermany:
CrLLz has already answered your question about the integration of the various streams and channels. Nevertheless, one more note and information about using a Dahua NVR with the Dahua integration: At least with my Dahua NVR (DHI-NVR2108-I), the Dahua integration "spamms" its protocol every second with log-ons and log-offs as soon as it is integrated into HA via it. This is also one of the reasons why I no longer use the Dahua integration for the NVR, but have integrated it via HA Onvif integration instead. This means that the NVR protocol is no longer "spammed".

This means that if you have integrated your NVR into HA via the Dahua integration, you should then take a look at the NVR protocol. ;-)

SaturnusDJ:
@CrLLz
Thanks, this indeed works. 👍🏼

@JimGermany
Agree because cam streams is not HA's core strength in any way.
I am most interested in all the metadata such as motion detection. Got this working nicely for a standalone IPC-T54IR-ZE-S3 but not for ANPR cams through the NVR. Soon multiple IPC-T54IR-ZE-S3 cams will be tested through NVR.

JimGermany:
> I am most interested in all the metadata such as motion detection

For me, the integration of the NVR via Onvif integration is sufficient, as all detection types (motion, field, line) are available as sensors under HA. Only the AI ​​features such as person and vehicle detection are missing.
[...]

PR #707, opened 2026-09-20
Say what the channel number means (fixes #646)

### LINKED ISSUE #646: Onvif
Hi there! 

Thanks for this integration, loving it!

I have one onvif camera connected to the NVR. I can't manage to get that stream. Tried all the channels. The standard onvif integration is not super stable. How can I add this kind of camera via this integration?

Kind regards

zios0007:
Closes #646.

The channel field reads:

> Channel (single cameras are 0, NVRs are the 0 based index)

That is accurate, and it still lets someone type the number their recorder shows — which is a different camera, or none at all.

Every Dahua NVR UI numbers its channels from 1, and so does the device. Measured on a DHI-NVR5464-16P-EI:

```
snapshot.cgi?channel=0   400 Bad Request   (21 bytes)
snapshot.cgi?channel=1   200               1,783,523 bytes
snapshot.cgi?channel=2   200               1,370,170 bytes
snapshot.cgi?channel=3   200                 886,951 bytes
```

The integration adds one to this field before asking, so **the recorder's channel 4 is 3 in the form**. Nothing says so, and "0 based index" is only obvious to someone who already knows.

### What that costs

#646 is a user who added two of their three cameras and could not find the third. They concluded the problem was that the third camera is an ONVIF camera attached to the NVR, and closed with *"Don't waste your time on this… Feel free to delete this issue."*

The integration never talks to that camera. It only ever talks to the recorder, which proxies every channel identically — there is no reference to ONVIF or `ProtocolType` anywhere in the package, and there is no code path on which a channel's upstream protocol could matter. So the one thing they were sure was the cause could not have been.

I cannot prove from here that an off-by-one is what they hit, and I have said so on the issue rather than asserting it. But it is the explanation that fits a user who can add some channels and not others, and the label is worth fixing either way.

### The change

```
Channel (a single camera is 0. On an NVR this is the 0-based index, which is
one less than the number the recorder shows: its channel 4 is 3 here)
```

Both forms are changed — the add and the reconfigure — and a test asserts they stay identical, because two different answers to the same question is worse than one.

Only `en.json` has this string; the other translations do not carry it and fall back.

### Testing

`tests/dahua/test_channel_label.py`: both forms exist, they match, and the label relates the index to what the recorder displays.

I kept the edit to the two label lines rather than reserialising the file, so the diff is 2 lines rather than the whole of `en.json`.


zios0007:
More evidence for this, found while looking at something else.

**#462 is the same confusion, independently, a year and a half earlier.** Someone adds an NVR5216-16P-EI and reports *"only camera 1 shows up"*. @CrLLz answers it correctly and has to spell out the very thing this label leaves implicit:

> channel 0 = channel 1 in the nvr
>
> And for exemple: Camera 4 will be channel 3.

That is the sentence the field should have contained. Two issues, two reporters, both stuck on the off-by-one, both resolved by a human explaining it in a comment — and #462 is still open.

So this is not a one-off misreading of a slightly terse label; it is the second recorded instance of the same failure, and the first one needed a bystander to rescue it.

Worth noting the asymmetry that makes it sting: the number the recorder shows you, the number in its own web UI, and the number in the RTSP URL you test with are **all 1-based**. This field is the only place in the workflow that is not, and it says so in four words at the end of a parenthesis.

rroller (approved):

PR #709, opened 2026-09-20
fix: deliver multipart events without heartbeat delay

kmich:
Follow-up to #678 and https://github.com/rroller/dahua/pull/678#issuecomment-5746992669.

#678 fixed ANPR event truncation by buffering multipart event data across TCP chunks, but the merged loop emits part N only after boundary N+1 arrives. With heartbeat=5, that can delay complete events such as DoorbellPressed until the next heartbeat.

This change keeps the buffering fix but uses each multipart part's Content-Length to emit the event as soon as its declared payload is complete. It keeps the next-boundary fallback for parts without Content-Length and preserves declared multipart boundaries.

Added regression tests proving a split multipart payload is delivered immediately after the final payload chunk, before any next boundary/heartbeat arrives.


zios0007:
@kmich this is the right fix and a better one than I would have written — reading `Content-Length` from the part's own headers, keeping the next-boundary path for devices that omit it, and honouring a declared boundary. Thank you for picking it up.

One robustness gap, with a reproduction against your actual function.

**A part whose declared `Content-Length` exceeds what the device sends stalls the stream permanently.** I extracted `_pop_complete_multipart_part` from this branch and ran it:

```
=== device declares Content-Length LARGER than it sends ===
   delivered: False
   -> STALLED although a next boundary is already in the buffer
   still stalled after two more whole parts arrive: True
```

Once `content_length` is set, the next-boundary fallback is skipped entirely:

```python
if content_length is not None:
    part_end = payload_start + content_length
    if len(buffer) < part_end:
        return None, buffer          # <- nothing else is tried
```

So the loop waits for bytes that never arrive, and every later event queues behind it. The previous code would have flushed that part on the next boundary.

**Why I do not think this is hypothetical.** #678 exists because payloads arrive split and, in the ANPR case, truncated. A truncated part *is* an under-delivery against its declared length — that is the same event, described from the other end. Whenever the device drops bytes mid-part rather than closing the connection, this now stalls instead of recovering at the next boundary.

The blast radius is the whole host: the event stream is shared by every channel on the device, so one bad part silences every camera on it until the read timeout forces a reconnect.

**The fix is small — prefer a boundary that is already in hand:**

```python
if content_length is not None:
    part_end = payload_start + content_length
    if len(buffer) >= part_end:
        part = buffer[:part_end]
        remainder = buffer[part_end:]
        ...
        return part, remainder
    # Declared more than arrived. If the next part has already started, the
    # length was wrong or the payload was truncated -- deliver what is framed
    # rather than waiting for bytes that are not coming.
    next_boundary = buffer.find(boundary, payload_start)
    if next_boundary == -1:
        return None, buffer
    return buffer[:next_boundary], buffer[next_boundary:]
```

That keeps every case your tests cover and makes a wrong length degrade to the old behaviour instead of a stall.

**One smaller note, same area.** The overflow guard now reads:

```python
if boundary not in buffer and len(buffer) > 131072:
    buffer = buffer[-4096:]
```

With the stall above, the boundary *is* in the buffer, so the trim never fires and the buffer grows without limit. The change above removes the stall and with it this consequence, so it probably needs nothing on its own — but it is worth knowing the guard no longer covers the case where a boundary is present and the part never completes.

**Worth cross-referencing:** #705 and #706 are open on the other end of this path. #705 fixes a truncated JSON payload raising `AttributeError` out of the event handler, and #706 stops any handler exception from leaving the read loop — which today takes the stream down for every channel on the host. They are complementary to this: yours reduces how often a part arrives incomplete, those stop an incomplete one from being fatal. No overlap in the diffs.

Happy to send the above as a patch against your branch if that is easier than taking it as a suggestion.

kmich:
> @kmich this is the right fix and a better one than I would have written — reading `Content-Length` from the part's own headers, keeping the next-boundary path for devices that omit it, and honouring a declared boundary. Thank you for picking it up.
> 
> One robustness gap, with a reproduction against your actual function.
> 
> **A part whose declared `Content-Length` exceeds what the device sends stalls the stream permanently.** I extracted `_pop_complete_multipart_part` from this branch and ran it:
> 
> ```
> === device declares Content-Length LARGER than it sends ===
>    delivered: False
>    -> STALLED although a next boundary is already in the buffer
>    still stalled after two more whole parts arrive: True
> ```
> 
> Once `content_length` is set, the next-boundary fallback is skipped entirely:
> 
> ```python
> if content_length is not None:
>     part_end = payload_start + content_length
>     if len(buffer) < part_end:
>         return None, buffer          # <- nothing else is tried
> ```
> 
> So the loop waits for bytes that never arrive, and every later event queues behind it. The previous code would have flushed that part on the next boundary.
> 
> **Why I do not think this is hypothetical.** #678 exists because payloads arrive split and, in the ANPR case, truncated. A truncated part _is_ an under-delivery against its declared length — that is the same event, described from the other end. Whenever the device drops bytes mid-part rather than closing the connection, this now stalls instead of recovering at the next boundary.
> 
> The blast radius is the whole host: the event stream is shared by every channel on the device, so one bad part silences every camera on it until the read timeout forces a reconnect.
> 
> **The fix is small — prefer a boundary that is already in hand:**
> 
> ```python
> if content_length is not None:
>     part_end = payload_start + content_length
>     if len(buffer) >= part_end:
>         part = buffer[:part_end]
>         remainder = buffer[part_end:]
>         ...
>         return part, remainder
>     # Declared more than arrived. If the next part has already started, the
>     # length was wrong or the payload was truncated -- deliver what is framed
>     # rather than waiting for bytes that are not coming.
>     next_boundary = buffer.find(boundary, payload_start)
>     if next_boundary == -1:
>         return None, buffer
>     return buffer[:next_boundary], buffer[next_boundary:]
> ```
> 
> That keeps every case your tests cover and makes a wrong length degrade to the old behaviour instead of a stall.
> 
> **One smaller note, same area.** The overflow guard now reads:
> 
> ```python
> if boundary not in buffer and len(buffer) > 131072:
>     buffer = buffer[-4096:]
> ```
> 
> With the stall above, the boundary _is_ in the buffer, so the trim never fires and the buffer grows without limit. The change above removes the stall and with it this consequence, so it probably needs nothing on its own — but it is worth knowing the guard no longer covers the case where a boundary is present and the part never completes.
> 
> **Worth cross-referencing:** #705 and #706 are open on the other end of this path. #705 fixes a truncated JSON payload raising `AttributeError` out of the event handler, and #706 stops any handler exception from leaving the read loop — which today takes the stream down for every channel on the host. They are complementary to this: yours reduces how often a part arrives incomplete, those stop an incomplete one from being fatal. No overlap in the diffs.
> 
> Happy to send the above as a patch against your branch if that is easier than taking it as a suggestion.

Thanks, you were right. I updated the branch to handle this.

The helper now checks for the next multipart boundary before trusting an unsatisfied declared payload end. If the next boundary is already present before `payload_start + Content-Length`, the framing wins and the part is delivered up to that boundary instead of waiting for bytes that are not coming.

I also added a regression test with an oversized `Content-Length` followed by a complete `DoorbellPressed` event, so the test verifies both that the malformed/truncated part is flushed and that the following event is still delivered.

I made the boundary check slightly stronger than the initial suggestion: it also covers the case where later events make the total buffer longer than the incorrect declared length, so we do not accidentally consume bytes from the next part just because the aggregate buffer became large enough.

Thanks for catching it.

zios0007:
Confirmed against the updated branch, and your version is better than what I suggested.

```
=== the stall ===
   oversized Content-Length, next boundary present -> delivered: True
   payload delivered:  b'Code=DoorbellPressed;action=Pulse;index=0'
   the FOLLOWING part still parses: b'Code=Heartbeat'

=== the case my patch got wrong ===
   oversized CL, later events make the buffer longer than part_end
   delivered:  b'Code=DoorbellPressed;action=Pulse;index=0'
   swallowed bytes from the next part? no
   next part intact: b'Code=VideoMotion;action=Start;index=0'

=== no regressions ===
   honest Content-Length, no next boundary -> delivered
   no Content-Length, boundary present     -> delivered
   no Content-Length, nothing following    -> held
```

**The strengthening you describe is the important part, and I had it wrong.** My suggestion tested `len(buffer) >= part_end` first, so once a couple of later events had arrived the aggregate buffer would satisfy an inflated `Content-Length` and the part would be delivered with the *next* event's bytes eaten into it — silently, and worse than the stall it was fixing, because a stall at least recovers on reconnect. Checking for the boundary before trusting an unsatisfied length is the right order and it holds in the test above.

Nothing further from me on this one. Thanks for turning it round so quickly.

PR #712, opened 2026-09-20
One table a device will not serve must not cost it RPC2 (#577)

zios0007:
Relevant to #577, and found by reading a live log rather than the source.

RPC2 exists so a poll costs one login instead of one per call (#636, #639). `_request` routes config reads over it and, on failure, decides whether to write the host off for the life of the process:

```python
if not rpc2_failure_is_permanent(rpc2_exception):
    ...                       # keep the transport
else:
    _HOST_RPC2_UNAVAILABLE.add(self._rpc2_key())
```

**A device that answers `result=false` for one table lands in the permanent branch.** `rpc2.request` raises a bare `ConnectionError` for that, and `TRANSIENT_RPC2_FAILURES` is `(TimeoutError, aiohttp.ClientConnectionError)`, so it does not match.

### Caught on live hardware

A DHI-NVR5464-16P-EI on 0.10.9. It serves RPC2 perfectly well and has no `LightingScheme` table — it answers `400 Bad Request` for that one over CGI too:

```
ConnectionError: Dahua RPC2 method configManager.getConfig returned result=false
(code=268959743, message=Unknown error! error code was not set in service!)

WARNING [custom_components.dahua] RPC2 config reads are not working for
192.168.0.213, using CGI instead
```

Then, hours later and repeatedly for the rest of the day:

```
WARNING Failed to sync device state for 192.168.0.213: 401, message='Unauthorized',
  url='https://192.168.0.213/cgi-bin/coaxialControlIO.cgi?action=getStatus&channel=1'
ERROR   Error fetching dahua data: 401, message='Unauthorized' ...
```

That is the login storm RPC2 was introduced to stop, back because one table was missing — and each occurrence fails the whole poll, so the entry goes unavailable.

**The device answering is the strongest evidence available that the transport works.** It was being read as proof that it does not.

### The change

`Rpc2MethodRefused` marks that case. It subclasses `ConnectionError`, so anything written against the old type still catches it. The refusal is recorded against **(host, table)** instead of the host: that table goes to CGI, everything else keeps the shared session.

It also **stops the retry**. `_rpc2_get_config` drops the shared session and logs in again on failure, because the failure it was written for is an expired session. Logging in again cannot change a device's mind about a table it does not have, so a refusal raises straight away rather than costing an extra login first.

### Simulated over one poll cycle plus a lighting read

Using the real `rpc2_failure_is_permanent` extracted from source:

```
BEFORE  RPC2: [MotionDetect]
        CGI : [LightingScheme, MotionDetect, Lighting_V2, LightingScheme,
               MotionDetect, VideoInOptions]          host written off: True

AFTER   RPC2: [MotionDetect, MotionDetect, Lighting_V2, MotionDetect,
               VideoInOptions]
        CGI : [LightingScheme, LightingScheme]        host written off: False
```

### Testing

`tests/dahua/test_rpc2_table_refusal.py`. The crux is the classification, so that is what the assertions pin: a refusal is a `ConnectionError` subclass, is **not** a transport failure, and — unchanged — a timeout still does not write the host off (#639) while an unrecognised failure still does.

### Note

This does not make the recorder serve `LightingScheme`; it has no such table and the `400` over CGI is correct behaviour for it. It stops that absence from being charged against every other read on the device.


zios0007:
@rroller a note on volume, since I have opened six PRs today and that is a lot to land on someone who mentioned being tied up with work. **Nothing here is urgent enough to deserve your evening**, but if you only look at two, these are the two:

**#712 (this one)** — one config table a device does not have writes RPC2 off for the whole host, so it falls back to a login per call. That is #577's login storm, and I caught it happening on my own recorder on 0.10.9: `LightingScheme` returns `result=false`, RPC2 is disabled, and the 401s start. 23 comments on #577 are people watching that in their NVR logs.

**#705, then #706** — a truncated event payload raises out of the event handler, and nothing catches it: `stream_events` wraps `on_receive` in `try`/`finally` with no handler, so it leaves the read loop and silences **every camera on that host** until the backoff reconnects. #705 fixes the one live instance (#475), #706 bounds the class. #706 contains #705's commit, so taking #706 alone is fine; taking #705 alone is also fine.

The other three are real but nobody is blocked on them, and they can sit as long as you like:

- **#707** — the channel field says "0 based index"; two reporters (#646, #462) have now been caught by it. Label text and a test.
- **#708** — the Amcrest doorbell's Security Light is a *select* reading a table only fetched for the *light* platform, so it reports Off forever if you disable lights. One gate, one row of behaviour changed.
- **#710** — warn when a channel is attached over ONVIF, because a recorder does not serve those on its Dahua paths and the camera silently 404s forever (#646).

**Two things that need you rather than me:**

**#670** by @ppgac has no CI checks at all, while every other open PR has three. It is their first PR here, so the workflows are waiting on **Approve and run workflows** from you — a rebase on their side will not trigger it. The fix itself closes #231 with a root cause and a regression test.

**#709** by @kmich removes the up-to-one-heartbeat delay on every event, which is worth having for doorbell presses. I reviewed it, found a case where a short part stalled the stream, and they had it fixed within the hour. It looks good to me now, for whatever an outside opinion is worth.

All six of mine are green and merge cleanly against main and each other, so order does not matter — except that #670 and #706 merge *cleanly* while leaving a gap, which I have flagged on #670.

